Issue #6, opened 2020-11-22
make the PassportJs strategy method and network agnostic


drgorb:
the reason we built the DID-Library is to allow for a method agnostic treatment of DIDs. The strategy that has been implemented here is not only method (ethr) but also network (Volta) specific.
If it is not possible to change this right now because the DID library is not mature enough, this should be a priority fix.

PR #11, opened 2020-12-09
Injecting ethers provider into LoginStrategy

jrhender:
@ChangoBuitrago and I were discussing that it could be useful for standard HTTP clients (e.g. Postman) to be able to test server applications which are using the `LoginStrategy`. In this case, it would be useful to be able to reuse a login claim, which isn't currently possible due to the checking of recent block number in claim (which is done to prevent replay attacks). However, if one was able to inject a test ethers provider (https://docs.ethers.io/v4/cookbook-providers.html#custom-provider) into the `LoginStrategy`, then the block number returned could be controlled and the claim could be reused.

Currently, this PR makes the change to the `LoginStrategy` to inject the provider and changes the express example to instantiate the RPCProvider (the nest example would need to be changed too)

I think this PR relates somewhat to issue #6 .

@ChangoBuitrago @manihagh @dwojno Any thoughts on this change? Is there a better way to allow the use "standard" HTTP clients in testing scenarios?

### DIFF
--- a/examples/server/express/package-lock.json
+++ b/examples/server/express/package-lock.json
@@ -13,6 +13,11 @@
         "negotiator": "0.6.2"
       }
     },
+    "aes-js": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/aes-js/-/aes-js-3.0.0.tgz",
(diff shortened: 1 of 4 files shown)


jrhender:
Closing for now, can revisit if functionality is needed in the future.

PR #24, opened 2021-01-20
[MYEN-545] add accepted roles check

dwojno:


jrhender (changes requested):
Nice work, I like the general approach 👍  However, please see PR #25 with my suggested changes and merge into this PR's branch if you approve, as I think there is a regression where null values are being included in returned claims.

dwojno:
@jrhender I think it will be easier to just fix the issue here, without merging branches :)

jrhender (approved):

PR #28, opened 2021-01-27
feat: add authorization for iam-cache-server

dwojno:


JGiter (approved):

lib/LoginStrategy.ts
```diff
@@ -1,15 +1,20 @@
...
 import axios, { AxiosInstance } from 'axios'
 import {
   ethrReg,
   Resolver,
   VoltaAddress1056,
 } from '@ew-did-registry/did-ethr-resolver'
 
-import { lookup, namehash, verifyClaim } from './utils'
+import {
+  createLoginTokenHeadersAndPayload,
```
jrhender, line 13: I think `utils` shouldn't really be the place for this function as, IMHO, this starts to move `utils` to be a bit of a function dumping ground. An idea for a refactor is that a `CacheServerClient` class is extracted. This would essentially be a wrapper around an axios client but could have methods like `loginToCacheServer` and `createLoginTokenHeadersAndPayload`. This allows a separation and a grouping of CacheServer functionality.

lib/LoginStrategy.ts
```diff
@@ -164,6 +182,32 @@ export class LoginStrategy extends BaseStrategy {
...
+    const sig = await this.signer.signMessage(arrayify(hash))
+    const encodedSignature = base64url(sig)
+    const claim = `${encodedHeader}.${encodedPayload}.${encodedSignature}`
+    const { data } = await this.httpClient.post<{ token: string }>('/login', {
+      claim,
+    })
+    this.httpClient.defaults.headers.common[
+      'Authorization'
+    ] = `Bearer ${data.token}`
+    console.log('DIDStrategy is now logged to cache server')
```
jrhender, line 208: Minor point but message might be a bit clearer if it said 'LoginStrategy is now logged in to cache server"

jrhender (approved):
Looks good to me, added some suggestions as well though

lib/LoginStrategy.ts
dwojno: Yeah, I was thinking about the same approach, let's do that 👍

dwojno: I've also added retry option when the response code is 401

PR #29, opened 2021-02-02
fix: endless loop when the cache server has auth strategy

dwojno:


lib/httpClient.ts
```diff
@@ -4,13 +4,15 @@ import { Signer, Wallet } from 'ethers'
...
 import { Claim, IRole } from './LoginStrategy.types'
+import { Policy } from 'cockatiel'
 
 export class CacheServerClient {
   private signer: Signer
   private httpClient: AxiosInstance
   private provider: JsonRpcProvider
   private failedRequests: Array<(token: string) => void> = []
   private isAlreadyFetchingAccessToken: boolean = false
+  public address: string
```
jrhender, line 15: I would set to be `readonly` (the `signer`, `provider` and `httpClient` should probably be as well)

lib/httpClient.ts
```diff
@@ -4,13 +4,15 @@ import { Signer, Wallet } from 'ethers'
 import { JsonRpcProvider } from 'ethers/providers'
 import { arrayify, keccak256 } from 'ethers/utils'
 import { Claim, IRole } from './LoginStrategy.types'
+import { Policy } from 'cockatiel'
```
jrhender, line 7: I'm not too familiar with retry policy libraries in javascript ecosystem, is cockatiel one of the most popular? (I know Polly is very popular in .NET, which cockatiel seems to be based off of)

lib/httpClient.ts
```diff
@@ -4,13 +4,15 @@ import { Signer, Wallet } from 'ethers'
...
 
 export class CacheServerClient {
```
jrhender, line 9: Would it be possible to rename this file to be "CacheServerClient.ts" to match the name of the class?

lib/LoginStrategy.ts
```diff
@@ -133,7 +135,8 @@ export class LoginStrategy extends BaseStrategy {
     }
 
     try {
-      const roleClaims = await this.getUserClaims(did)
+      const roleClaims =
+        this.strategyAddress === address ? [] : await this.getUserClaims(did)
```
jrhender, line 139: IMO, a comment here to explain why this is needed would be helpful to future readers. E.g. "getUserClaims attempts to retrieve claims from cache-server and so when the cache-server itself is using the LoginStrategy, this creates a login attempt loop. Therefore, not getting userClaims if address attempting to login is the address of the strategy." (<- Feel free to correct that if it's wrong)

jrhender (approved):
Looks good, added a few suggestions

lib/httpClient.ts
dwojno: 👍

lib/httpClient.ts
dwojno: Yeah it's pretty much popular, we're using at TSH this library for production ready apps :)

lib/httpClient.ts
dwojno: 👍

lib/LoginStrategy.ts
dwojno: Great idea, I'm forgetting to comment that kind of stuff :) no it says everything so I'll use it :)

PR #31, opened 2021-02-08
feat(jwt): add option to not generate jwt inside strategy

dwojno:
I created this PR, to add option to not create a jwt token inside the strategy. It may be useful when somebody just wants to authenticate user and handle their own way of authorizing.
If the token is not created, the strategy will assign to `req.user`:
```
did: string;
verfiedRoles: string[]
```
This PR does not contain any breaking changes.

### DIFF
--- a/lib/LoginStrategy.ts
+++ b/lib/LoginStrategy.ts
@@ -39,7 +39,7 @@ interface LoginStrategyOptions extends StrategyOptions {
 
 export class LoginStrategy extends BaseStrategy {
   private readonly claimField: string
-  private readonly jwtSecret: string | Buffer
+  private readonly jwtSecret?: string | Buffer
   private readonly jwtSignOptions?: jwt.SignOptions
   private readonly provider: providers.JsonRpcProvider
   private readonly cacheServerClient: CacheServerClient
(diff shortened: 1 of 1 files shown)


jrhender:
Seems like a reasonable idea. Did someone request this functionality?

jrhender (commented):
I think the `@ApiBody` documentation in the `LoginController` needs to be updated now as an "identitytoken" isn't necessarily being returned, right?

dwojno:
> I think the `@ApiBody` documentation in the `LoginController` needs to be updated now as an "identitytoken" isn't necessarily being returned, right?

The strategy by default is returning signed JWT Token, and yes the login controller does not need to do it, but the strategy is not returning the token by itself, it is just passing the token in `request.user` and sending back the token to user is responsibility of the controller.

In our case we we'll still generate the JWT token in `/login` route, but not inside the Strategy.

I think this approach is more flexible cause maybe someone don't like JWT token, and want to use strategy just for validation the user and handle authentication differently, before this PR that wasn't possible :)

jrhender:
> > I think the `@ApiBody` documentation in the `LoginController` needs to be updated now as an "identitytoken" isn't necessarily being returned, right?
> 
> The strategy by default is returning signed JWT Token, and yes the login controller does not need to do it, but the strategy is not returning the token by itself, it is just passing the token in `request.user` and sending back the token to user is responsibility of the controller.
> 
> In our case we we'll still generate the JWT token in `/login` route, but not inside the Strategy.
> 
> I think this approach is more flexible cause maybe someone don't like JWT token, and want to use strategy just for validation the user and handle authentication differently, before this PR that wasn't possible :)

Ah yea, sorry, silly suggestion on my part, the LoginController is a cache-server thing and this is for passport-did-auth 🤦 . Still I think we should document this option somewhere. I guess in this case maybe it would go in the class documentation for `LoginStrategy`?

jrhender (approved):

PR #36, opened 2021-02-23
v1.0 release

jrhender:
Merge from develop to master to release v1.0


dwojno (approved):

jrhender:
Decided to release so that Messaging and EWAI can begin to integrate

jrhender:
Actually, thinking about it some more, it would be confusing to users of passport-did-auth if we have a v1.0 that doesn't currently work with volta-identitycache.energyweb.org. So I will change back to draft for now 😅

PR #42, opened 2021-05-27
Feature/swtch 866 authenticate based on did document

knzeng-e:
#### Pull request Content:

* Update: In `LoginStrategy.ts`, I replaced the `verifyClaim` function (imported from `utils.ts`) by `verifyDidClaim` function. 
This new function is coded as a method of LoginStrategy class and checks into the `authentication` field of the DID to validate the issuer.

* Add: A `LoginStrategy.test.ts` file containing unit tests of the new `verifyDidClaim` method
* Some light reformating of `LoginStrategy.ts` code for a better readibility


lib/utils.js
```diff
@@ -0,0 +1,104 @@
+"use strict";
```
jrhender, line 1: Can you explain this `utils.js` file? Was this something you wrote or is it compiled javascript that you copied in from another source?

lib/utils.ts
```diff
@@ -87,3 +89,7 @@ export const verifyClaim = (token: string, { iss }: ITokenPayload) => {
   const addressFromDigest = recoverAddress(digest, signature)
   return decodedAddress === addressFromDigest ? iss : ''
 }
+
+export const decodeJwToken = (token: string) : any => {
```
jrhender, line 93: Is there a `@types` package for `jsonwebtoken` so that the specific type can be specified instead of returning `any`?

lib/LoginStrategy.ts
```diff
@@ -239,4 +252,34 @@ export class LoginStrategy extends BaseStrategy {
...
+    const claimInfos: any = decodeJwToken(claimedToken)
+
+    //extract all authenticated users and filter to keep those owning the claimed DID
+    const authenticated = didDocument.authentication.filter(auth => {
+      auth["publicKey"] === `${claimInfos.did}#owner`.toString()
+    })
+
+    return authenticated.length === 0
+  }
+}
```
jrhender, line 285: A solid first attempt 👍  but not quite correct

So our DID documents, entries in `authentication` property provide links to entries in the `publicKey` property. So our current DID documents are like this:
```
{
  "id": "did:ethr:0x96142FAC151454E38B897cE50fdE26C60622B14E",
  "service": [],
  "authentication": [
    {
      "type": "owner",
      "validity": {
        "_hex": "0x1fffffffffffff"
      },
      "publicKey": "did:ethr:0x96142FAC151454E38B897cE50fdE26C60622B14E#owner"
    }
  ],
  "publicKey": [
    {
      "id": "did:ethr:0x96142FAC151454E38B897cE50fdE26C60622B14E#key-owner",
      "type": "Secp256k1veriKey",
      "controller": "0x96142FAC151454E38B897cE50fdE26C60622B14E",
      "publicKeyHex": "0x035f58d27f27944f42089083a6c96b099360bb6099988bcfe3ddca74541c0a097d"
    }
  ]
}
```
So in this example, `"publicKey": "did:ethr:0x96142FAC151454E38B897cE50fdE26C60622B14E#owner"` in the `authentication` section links to the key with `"id": "did:ethr:0x96142FAC151454E38B897cE50fdE26C60622B14E#key-owner"` in the `publicKey` property (both should end with `#key-owner`, this is a bug in ew-did-registry that you should account for. The actual public key value that should be used to verify the signature though is `"publicKeyHex": "0x035f58d27f27944f42089083a6c96b099360bb6099988bcfe3ddca74541c0a097d"`.

In other words, the `publicKey` `id` is arbitrary. The code should be able to verify a signature using a DID document like this:
```
{
  "id": "did:ethr:0x96142FAC151454E38B897cE50fdE26C60622B14E",
  "service": [],
  "authentication": [
    {
      "type": "owner",
      "validity": {
        "_hex": "0x1fffffffffffff"
      },
      "publicKey": "my-favorite-public-key-1"
    }
  ],
  "publicKey": [
    {
      "id": "my-favorite-public-key-1",
      "type": "Secp256k1veriKey",
      "controller": "0x96142FAC151454E38B897cE50fdE26C60622B14E",
      "publicKeyHex": "0x035f58d27f27944f42089083a6c96b099360bb6099988bcfe3ddca74541c0a097d"
    }
  ]
}
```

jrhender (changes requested):
Congrats on the first PR 🎉

lib/utils.js
knzeng-e: This file was automatically generated at the `utils.ts` compilation. Since the former `verifyClaim` function was declared inside it and then imported in `LoginStrategy.ts` file, the corresponding .js file as been compiled when I made changes to it.

jrhender: Is it necessary to include in the project? Can you import from the `utils.ts` file?

lib/utils.ts
knzeng-e: Yes ok, I wiil fix it. The returned type is `string | { [key: string]: any; }`

lib/LoginStrategy.ts
knzeng-e: That Makes a lot of sens, and gets clearer in my mind! Thank you for your careful review and your precious explanations John. I'm on it !

lib/utils.js
knzeng-e: Since those .js files are just build on ts files, it is not necessary indeed to include in the project.

lib/LoginStrategy.ts
```diff
@@ -239,4 +253,49 @@ export class LoginStrategy extends BaseStrategy {
...
+   * @description checks if the claimer is authenticated into the DID Document
+   *
+   * @param token
+   * @param payload
+   *
+   * @returns {string} issuer DID or null
+   */
+  async verifyDidClaim(token: string, payload: ITokenPayload): Promise<string> {
+
+    const didDocument = await this.didResolver.read(payload.iss)
```
jrhender, line 267: Better to use the `cacheServer` if it is available as it will potentially be much faster.
```suggestion
    const didDocument = await this.cacheServerClient?.getDidDocument(payload.iss) ??  await this.didResolver.read(payload.iss)
```

lib/LoginStrategy.ts
```diff
@@ -239,4 +253,49 @@ export class LoginStrategy extends BaseStrategy {
...
+
+    if (this.isAuthorized(token, didDocument))
+      return payload.iss
+    return null 
+  }
+
+  private isAuthorized = (claimedToken: string,didDocument: IDIDDocument) : boolean => {
```
jrhender, line 274: Nice extraction of this `isAuthorized` function 👍

lib/LoginStrategy.ts
```diff
@@ -239,4 +253,49 @@ export class LoginStrategy extends BaseStrategy {
...
+
+    //retrieve claim infos from token
+    const claimInfos = jwt.decode(claimedToken) as {[key: string]: any;}
```
jrhender, line 277: I think it would be better to install the `@types/jsonwebtoken` package and then you can have an known return type from `jwt.decode` https://www.npmjs.com/package/@types/jsonwebtoken

lib/LoginStrategy.ts
```diff
@@ -239,4 +253,49 @@ export class LoginStrategy extends BaseStrategy {
...
+    const didPubKey = didDocument.publicKey
+
+    if (didPubKey.length === 0)
+      return false
+    
+    //extract all authenticated users and keep those refering issuer's publicKey
+    const publicKeys = didDocument.authentication.map((auth, index) => {
+      const auth_pubkeyRef = auth["publicKey"] //public key reference in authentication field
+      const pubkey_id = didPubKey[index]["id"] //public key Id in publickKey field of DID document
```
jrhender, line 286: I don't think `index` can be used as it is currently being used. I think it would fail for a DID document like this one, right?
```
  "authentication": [
    {
      "publicKey": "key-2"
    }
  ],
  "publicKey": [
    {
      "id": "key-1",
      "publicKeyHex": "<some public key>"
    },
    {
      "id": "key-2",
      "publicKeyHex": "<some public key>"
    }
  ]
```

lib/LoginStrategy.ts
```diff
@@ -239,4 +253,49 @@ export class LoginStrategy extends BaseStrategy {
...
+
+      /* If the auth field refers to the this public key entry, 
+        get the real public key in the publicKeyHex field */
+
+      if (extractId(auth_pubkeyRef) === extractId(pubkey_id)){
```
jrhender, line 291: `extractId` function isn't needed. The `id` is the entire string, not just the tag after the `#`

lib/LoginStrategy.ts
```diff
@@ -239,4 +253,49 @@ export class LoginStrategy extends BaseStrategy {
...
+        return didPubKey[0]["publicKeyHex"] //get the real public key in DID document
```
jrhender, line 292: Can you explain why `0` here?

lib/LoginStrategy.ts
```diff
@@ -239,4 +253,49 @@ export class LoginStrategy extends BaseStrategy {
...
+      }
+    })
+
+    //Compute the address based on the returned public Key and compare it to the address
+    // encoding the did
+    const isPublicKeyValid = (computeAddress(publicKeys[0]) === claimInfos.iss.split(':')[2])
```
jrhender, line 298: This comparison isn't adequate, I believe. The provider of the token could simply lie and put an `iss` that doesn't match the signature.

Instead, something like this is necessary to actually verify the signature:
(this is code that can be imported from the ew-did-registry `jwt` package)
```
const possibleKeys = (message: string, signature: string): string[] => [
  arrayify(keccak256(message)),
  arrayify(hashMessage(arrayify(keccak256(message)))),
]
  .map((h) => recoverPublicKey(h, signature))
  .map((key) => computePublicKey(key, true).slice(2));

const verifyEthersSignature = () => async (token: string, publicKey: string):
  Promise<{ success: boolean; payload?: unknown }> => {
  const [encodedHeader, encodedPayload, encodedSignature] = token.split('.');
  const msg = `0x${Buffer.from(`${encodedHeader}.${encodedPayload}`).toString('hex')}`;
  const signature = base64url.decode(encodedSignature);

  // eslint-disable-next-line no-restricted-syntax
  for (const key of possibleKeys(msg, signature)) {
    if (key === publicKey) {
      return { success: true, payload: JSON.parse(base64url.decode(encodedPayload)) };
    }
  }
  return { success: false, payload: {} };
};
```

lib/LoginStrategy.ts
jrhender, line 274: I think there are some of good pieces here, but I would suggest the following algorithm for this function:

1. filter the DID document public keys for all of the keys that are also in the `authentication` section of the document. This provides the "authentication" publicKeys
2. For each authentication publicKey, run the `verifyEthersSignature` function from the ew-did-registry `jwt` package to see if it matches. If one of the publicKeys verifies, then return true.

test/LoginStrategy.test.ts
```diff
@@ -0,0 +1,90 @@
...
+const keys = new Keys({
+    privateKey: process.env.PRIVATE_KEY,
+  });
+  const secondKeys = new Keys({
+    privateKey: process.env.PRIVATE_KEY_2,
+  });
+  const issuerDID = `did:ethr:${keys.getAddress()}`
+  const secondDID = `did:ethr:${secondKeys.getAddress()}`
+
+describe("Testing LoginStrategy", () => {
```
jrhender, line 44: Just a heads up, I think that the tests are good but as they are against deployed infrastructure (e.g. the volta RPC), I'm not sure they can be run during CI builds as deployed infrastructure may not always be available.

jrhender (changes requested):
Some more comments. Keep working at it, making process 👍

lib/LoginStrategy.ts
knzeng-e: For this I had a concern about the fact that the reference in the authentication field differs, after the '#' character, from the one in publickey field. So for now I extracted the parts before the #, which are similar in both fields.

test/LoginStrategy.test.ts
jrhender: @knzeng-e we can disable for ci by adding an `x` in front of the `describe`
```suggestion
xdescribe("Testing LoginStrategy", () => {
```

lib/utils.ts
```diff
@@ -87,3 +94,14 @@ export const verifyClaim = (token: string, { iss }: ITokenPayload) => {
   const addressFromDigest = recoverAddress(digest, signature)
   return decodedAddress === addressFromDigest ? iss : ''
 }
+
+//used to compare ids in DID Since the referenced pubkey id differs slighty from the auth reference Id
+export const areLinked = (authId :string, pubKeyID: string)  =>  {
```
jrhender, line 99: Just curious, why put this in `utils.ts`?

lib/LoginStrategy.ts
```diff
@@ -239,4 +255,50 @@ export class LoginStrategy extends BaseStrategy {
...
+
+    //get all authenticated public keys
+    const authenticatedPubkeys = didDocument.publicKey.filter(pubkey => {
+      return this.isAuthenticated(pubkey, didDocument.authentication)
+    })
+
+    //Keep in an array all keys passing the verifying process
+    const arekeysValid = authenticatedPubkeys.filter(async (key) => {
+      const publicKey = key["publicKeyHex"]
+      const keyStatus = await verifyEthersSignature(claimedToken, publicKey)
```
jrhender, line 291: Why using `verifyEtherSignature` directly instead of `JWT` class? (Not saying it's wrong necessarily, just trying to understand 😃 )

jrhender (commented):
Hi Kevin, didn't get a chance to fully review but will look at it some more tomorrow

lib/LoginStrategy.ts
jrhender: If possible, I think it is slightly better to use the `JWT` class as that is the one that is exported cleanly from the `ew-did-registry/jwt` package. Not a big deal though.

test/LoginStrategy.test.ts
jrhender, line 44: A suggestion that I have is that you refactor the code that you have written into a separate class (similar to the current `ClaimVerifier`). In this way, you can write unit tests against the class that can test the various scenarios without needing the blockchain. In will also have the advantage of encapsulating your functions more directly. For example, you could have some class like:
```
AuthenticationTokenVerifier{
  public verify(token, didDocument) {
     ...
  }
  .... all of your private methods
}
```
In this way, you can inject various token and didDocument combinations into the `verify` function to unit test. Does this make sense?

lib/LoginStrategy.ts
```diff
@@ -1,56 +1,71 @@
...
+  areLinked,
+} from './utils'
 
-import { lookup, namehash, verifyClaim } from './utils'
 import {
   Claim,
-  DecodedToken,
   IRole,
-  IRoleDefinition,
+  DecodedToken,
```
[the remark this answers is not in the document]
jrhender: Perhaps a general comment related to is that if you haven't already, if you're using VS code, you can install the ES Lint extension to helpful point out possibly improvements like this.

lib/LoginStrategy.ts
jrhender: Yes, you are right. If key is in the `publicKey` array but not in the `authentication` array, then my understanding is that it is not a valid `authentication` key. 

That said, there is another scenario that I haven't brought up yet that we should accommodate 😅 
Basically, I think you've covered this scenario, where an authentication claim signed by a key like this one should pass:
```
{
  "id": "did:ethr:0x96142FAC151454E38B897cE50fdE26C60622B14E",
  "service": [],
  "authentication": [
    {
      "type": "owner",
      "validity": {
        "_hex": "0x1fffffffffffff"
      },
      "publicKey": "my-favorite-public-key-1"
    }
  ],
  "publicKey": [
    {
      "id": "my-favorite-public-key-1",
      "type": "Secp256k1veriKey",
      "controller": "0x96142FAC151454E38B897cE50fdE26C60622B14E",
      "publicKeyHex": "0x035f58d27f27944f42089083a6c96b099360bb6099988bcfe3ddca74541c0a097d"
    }
  ]
}
```

There is also a second scenario:
```
{
  "id": "did:ethr:0x96142FAC151454E38B897cE50fdE26C60622B14E",
  "service": [],
  "authentication": [],
  "publicKey": [
    {
      "id": "my-favorite-public-key-1",
      "type": "Secp256k1sigAuth",
      "controller": "0x96142FAC151454E38B897cE50fdE26C60622B14E",
      "publicKeyHex": "0x035f58d27f27944f42089083a6c96b099360bb6099988bcfe3ddca74541c0a097d"
    }
  ]
}
```
Basically, this key isn't listed in the `authentication` array but because it is the `sigAuth` type, it is valid authentication key.

(@JGiter I am correct in saying these two scenarios are what should constitute valid "authentication" keys?)

jrhender: @JGiter I have been wondering about this because our current DID document "creation" gives the key a `veriKey` type but then the key is added to the `authentication` array, which is contrary to what I described above... 🤔

lib/LoginStrategy.ts
```diff
@@ -1,56 +1,71 @@
...
-import { PublicResolver } from '../ethers/PublicResolver'
+
+import {
+  ethrReg,
+  Resolver,
+  VoltaAddress1056,
+} from '@ew-did-registry/did-ethr-resolver'
+
+import { verificationMethods } from '@ew-did-registry/jwt/dist/verify'
+import { JWT } from '@ew-did-registry/jwt'
```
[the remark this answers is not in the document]
knzeng-e: yes it's true. In fact I should have put my PR as a draft, since I normally need the JWT class to use one of its verify functions. But I was wondering if I could avoid creating a hole instance and just use directly the desired function, and I went exploring this second path. I'll clean all that.

lib/LoginStrategy.ts
knzeng-e: Very interesting, thank you John for pointing those different scenarios. Indeed, I already cover the scenario where a claimed is signed with that kind of key:
`"publicKey": [
    {
      "id": "my-favorite-public-key-1",
      "type": "Secp256k1veriKey",
      "controller": "0x96142FAC151454E38B897cE50fdE26C60622B14E",
      "publicKeyHex": "0x035f58d27f27944f42089083a6c96b099360bb6099988bcfe3ddca74541c0a097d"
    }
  ]`
  
  So I targuet on managing the scenario with a sigAuth type of key

jrhender: Found the mapping in the current DIF did-ethr-resolver that handles this case: https://github.com/decentralized-identity/ethr-did-resolver/blob/18f4488b2bffe8cd3939224020b386cecfea9c9f/src/resolver.ts#L215

If `sigAuth` is used, then they are adding to the `auth` array. So I think we are on the correct path.

lib/utils.ts
knzeng-e: I used this pattern thinking that this function could be needed in other classes. But yes, I agree with your views about "utils"

lib/LoginStrategy.ts
```diff
@@ -239,4 +254,78 @@ export class LoginStrategy extends BaseStrategy {
...
+
+  /**
+   * @description checks if the claimer is authenticated into the DID Document
+   *
+   * @param token
+   * @param payload
+   *
+   * @returns {string} issuer DID or null
+   */
+  async verifyClaim(token: string, payload: ITokenPayload): Promise<string> {
```
jrhender, line 266:
```suggestion
  async verifyAuthenticationClaim(token: string, issuerDID: string): Promise<string> {
```
To me this is a clearer function signature as it's not the entire `ITokenPayload` that we need, it is just the `payload.iss` so better to no force calls to provide an entire `ITokenPayload`.
If this change, is made then the description of the function could be changed to "checks a token was signed by the issuer DID or a valid authentication delegate of the issuer"

lib/LoginStrategy.ts
```diff
@@ -239,4 +254,78 @@ export class LoginStrategy extends BaseStrategy {
...
+    if (didPubKey.length === 0)
+      return false
+
+    const keys = new Keys({privateKey: this.privateKey})
+    const jwtSigner =  new JWT(keys)
+
+    //get all authenticated public keys
+    const authenticatedPubkeys = didDocument.publicKey.filter(pubkey => {
+      return this.isAuthenticated(pubkey, didDocument.authentication)
+    })
```
jrhender, line 286: A bit of a minor point, but I think it's clearer if the wording used is "authentication" public keys. This is because the keys are used for authentication, the key are not themselves authenticated.
```suggestion
    //get all authentication public keys
    const authenticationPubkeys = didDocument.publicKey.filter(pubkey => {
      return this.isAuthenticationKey(pubkey, didDocument.authentication)
    })
```

lib/LoginStrategy.ts
```diff
@@ -239,4 +254,78 @@ export class LoginStrategy extends BaseStrategy {
...
+  private filterValidKeys = async (authenticatedKey: IPublicKey[], verifSignature) => {
+    const results = await Promise.all(authenticatedKey.map(verifSignature));
+
+    return authenticatedKey.filter((_key, index) => results[index]);
+  }
+
+  private isAuthenticated = (publicKey: IPublicKey, authFieldDocument: (string | IAuthentication)[]) => {
+    const authenticatedKeys = authFieldDocument.map(auth => {
+      if (auth.length === 0 && publicKey !== undefined)
+       return this.isSigAuth(publicKey["type"])
```
jrhender, line 309: I think this might be an issue. I think that the `isSigAuth` case is applicable even if `auth.length !== 0`

jrhender (commented):
This is looking good! I think if we can get the tests to not depend on external dependencies then we should be in good shape!

lib/LoginStrategy.ts
knzeng-e: Yes, the way I think about this is have a particular look on sigAuth type when we have the case where authentication field is empty. But if `auth.length !== 0` it will proceed with the normal bind checking using the `areLinked` function. Thus, a `sigauth` publickey type is handled in this scenario the sameway as a `verikey`.  Am I missing something ?

jrhender: This is an example of a case with two valid "authentication" keys
```
{
  "id": "did:ethr:0x96142FAC151454E38B897cE50fdE26C60622B14E",
  "service": [],
  "authentication": [
    {
      "type": "owner",
      "validity": {
        "_hex": "0x1fffffffffffff"
      },
      "publicKey": "my-favorite-public-key-1"
    }
  ],
  "publicKey": [
    {
      "id": "my-favorite-public-key-1",
      "type": "Secp256k1veriKey",
      "controller": "0x96142FAC151454E38B897cE50fdE26C60622B14E",
      "publicKeyHex": "0x035f58d27f27944f42089083a6c96b099360bb6099988bcfe3ddca74541c0a097d"
    },
    {
      "id": "my-favorite-public-key-1",
      "type": "Secp256k1sigAuth",
      "controller": "0x96142FAC151454E38B897cE50fdE26C60622B14E",
      "publicKeyHex": "0x035f58d27f27944f42089083a6c96b099360bb6099988bcfe3ddca74541c0a097d"
    }
  ]
}
```

test/TestDidDocuments.ts
```diff
@@ -0,0 +1,139 @@
...
+    ],
+    created: null,
+    delegates: null,
+    proof: null,
+    publicKey: [
+      {
+        id: 'did:ethr:0x72EFf9faB7876c4c1b6cAe426c121358431758F3#key-owner',
+        type: 'Secp256k1veriKey',
+        controller: '0x72EFf9faB7876c4c1b6cAe426c121358431758F3',
+        publicKeyHex: '0x0355f6f95a109a1ebe2a2aaea74f49274ea55aaecd2d32a44a2002bcf63002b913'
```
jrhender, line 49: IMO, would be better instead of the string, if the variable was used so that it was clear to a reader of the code that the publicKey in this DID doc is from the private key that signed some message. (I might not have put this in the right spot, but hopefully it is clear 😄 )
```suggestion
        publicKeyHex: firstKeys.publicKey
```

jrhender (commented):
Nice, this is looking really good, I think we are almost there!

test/TestDidDocuments.ts
```diff
@@ -0,0 +1,92 @@
...
+
+dotenv.config();
+
+const firstKeys = new Keys({
+  privateKey: process.env.PRIVATE_KEY
+});
+
+const secondKeys = new Keys({
+  privateKey: process.env.PRIVATE_KEY_2
+});
```
jrhender, line 13: Similar to the above, I think it is easier just to generate keys. Then we don't need to have environment variables in the automated tests and developers don't need to setup environment variables.
```suggestion
import { Keys } from '@ew-did-registry/keys'
import {IDIDDocument} from "@ew-did-registry/did-resolver-interface"

const firstKeys = new Keys();

const secondKeys = new Keys();
```

jrhender (commented):
Just a comment about using private keys from env file. Let me know what you think of the suggestion 😄

knzeng-e:
The fact is that I will have to generate those keys both in the main testing file `AuthTokenVerifier.test` and the `TestDidDocuments` file. Since I need to have the same keys in both files, generating keys each time we run tests would lead to different keys. 
But I aggree with you, it would just be better if developpers don't have to set private keys in an .env file in order to run those tests. 
Maybe we can hardcode some testing private keys ?

knzeng-e (commented):
> The fact is that I will have to generate those keys both in the main testing file `AuthTokenVerifier.test` and the `TestDidDocuments` file. Since I need to have the same keys in both files, generating keys each time we run tests would lead to different keys.
> But I aggree with you, it would just be better if developpers don't have to set private keys in an .env file in order to run those tests.
> Maybe we can hardcode some testing private keys ?

Or maybe I could just export generated keys from `TestDidDocuments` to `AuthTokenVerifier.test` .. humm

jrhender:
@knzeng-e I think exporting the generated keys from `TestDidDocuments` to `AuthTokenVerifier.test` is a good idea 👍

knzeng-e:
> @knzeng-e I think exporting the generated keys from `TestDidDocuments` to `AuthTokenVerifier.test` is a good idea

@jrhender That's done ! Thank you, the code has a cleaner look and feel now  !

package.json
```diff
@@ -56,6 +56,7 @@
     "@typescript-eslint/eslint-plugin": "^4.15.1",
     "@typescript-eslint/parser": "^4.15.1",
     "babel-jest": "^26.6.3",
+    "dotenv": "^10.0.0",
```
jrhender, line 59: Please remove from package.json if not used
```suggestion
```

jrhender (approved):
Just the one package to remove but I think we're good to go 👍

PR #48, opened 2021-06-25
Fix/swtch 621 passport did auth compile in strict mode

ahmedolaibrahim:
- [x] Fix errors in compilation mode


knzeng-e (approved):
Sounds good @ahmedolaibrahim !

PR #50, opened 2021-06-25
1.1.0 Release

jrhender:
I believe changes are:
- update to ew-did-registry dependency to fix ethers issues
- implementation of auth verification based on DID Document keys


knzeng-e (approved):
So cool !

PR #52, opened 2021-07-02
[bug/SWTCH-1091] Adds build:api_docs to deploy process

ahmedolaibrahim:
- [x] Adds build:api_docs to deploy workflow


jrhender:
The travis build is failing. do you know why this is?
image

jrhender (commented):
I don't think it will be sufficient to simply run `npm run build:api_docs` during build as (at least currently), the docs need to be committed to source control. So maybe the commit needs to be made automatically? Or maybe there needs to be a check that fails the build if the documentation hasn't been updated by the developer? Or maybe you can come up with some better idea.

ahmedolaibrahim:
> I don't think it will be sufficient to simply run `npm run build:api_docs` during build as (at least currently), the docs need to be committed to source control. So maybe the commit needs to be made automatically? Or maybe there needs to be a check that fails the build if the documentation hasn't been updated by the developer? Or maybe you can come up with some better idea.

I am not quite sure why the build is failing but i'll check that.  I think you are right about the build not being committed to source control luckily I found an article that talks about how to do that [here](https://medium.com/@russleyshaw/typescript-package-deployment-with-travisci-df788ffb8563)

jrhender:
> I think you are right about the build not being committed to source control luckily I found an article that talks about how to do that [here](https://medium.com/@russleyshaw/typescript-package-deployment-with-travisci-df788ffb8563)

@ahmedolaibrahim Nice research 👍  . I took a look at the article though and I think it is not committing the docs to source control (which we need for read the docs), it is only bundling the updated docs into the NPM package. So I'm not sure it will work in our case? 🤔

package.json
```diff
@@ -13,7 +13,7 @@
     "build:ts": "tsc",
     "build:contracts": "npm run build:typechain:resolver",
     "build:typechain:resolver": "typechain --target ethers-v4 --outDir ethers './node_modules/@ensdomains/resolver/build/contracts/PublicResolver.json'",
-    "build:api_docs": "rm -rf docs/api && typedoc",
+    "build:api_docs": "rm -rf docs/api && typedoc --plugin typedoc-plugin-markdown --hideBreadcrumbs true",
```
ahmedolaibrahim, line 16: asides rendering the documentation as markdown, I also added this so we have consistency across our repos as this is also used by `iam-client-lib`

typedoc.json
```diff
@@ -1,11 +1,9 @@
 {
-  "inputFiles": ["lib/"],
+  "entryPoints": ["lib/"],
```
ahmedolaibrahim, line 2: these options have since been removed in version 0.20 as stated in the release [notes](https://github.com/TypeStrong/typedoc/releases/tag/v0.20.0)

jrhender: Nice catch

jrhender (approved):
Looks good to me, nice work!

PR #102, opened 2021-09-20
feat(LoginStrategy): add cacheServerVersion param

jrhender:
This param is append to the end of the cacheServerUrl if present.

### DIFF
--- a/lib/LoginStrategy.ts
+++ b/lib/LoginStrategy.ts
@@ -28,6 +28,11 @@ interface LoginStrategyOptions extends StrategyOptions {
   claimField?: string
   rpcUrl: string
   cacheServerUrl?: string
+  /** 
+   * Provides the version string to be appended to the cacheServerUrl.
+   * For example: https://volta-identitycache.energyweb.org/v{cacheServerVersion}
+   */
+  cacheServerVersion?: string
   privateKey?: string
   numberOfBlocksBack?: number
   ensResolverAddress?: string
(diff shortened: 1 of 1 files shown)


ahmedolaibrahim (approved):
Looks good to me..

jrhender:
@ahmedolaibrahim I'm a bit confused why this would be necessary actually. It seems one could accomplish the same thing with the current version of `passpor-did-auth` by simply passing in the full URL. E.g. instead of using `cacheServerUrl: https://identitycache-dev.energyweb.org/`, use `cacheServerUrl: https://identitycache-dev.energyweb.org/v1/`

JGiter (approved):
Have you tried if this works? I tried this way by appending `v1` to cacheServerUrl in `AuthStrategy` constructor on cache server but had 401

ahmedolaibrahim:
> @ahmedolaibrahim I'm a bit confused why this would be necessary actually. It seems one could accomplish the same thing with the current version of `passpor-did-auth` by simply passing in the full URL. E.g. instead of using `cacheServerUrl: https://identitycache-dev.energyweb.org/`, use `cacheServerUrl: https://identitycache-dev.energyweb.org/v1/`

@jrhender my argument against adding versions as part of the base URL is that a user of this library might want to use v1 for some routes on the cache server and maybe another version e.g `/v2` depending on their needs. adding it as part of a base URL defeats that purpose.

jrhender:
Closing as not necessary to accomplish https://github.com/energywebfoundation/iam-cache-server/pull/162

PR #104, opened 2021-09-22
fix(cacheServer): fix init cache server client

Harasz:


lib/LoginStrategy.ts
```diff
@@ -43,14 +43,14 @@ export class LoginStrategy extends BaseStrategy {
...
   private readonly provider: providers.JsonRpcProvider
-  private readonly cacheServerClient!: CacheServerClient
   private readonly numberOfBlocksBack: number
   private readonly ensResolver: PublicResolver
   private readonly didResolver: Resolver
   private readonly ipfsStore: DidStore
   private readonly acceptedRoles: Set<string>
   private readonly strategyAddress?: string
   private readonly privateKey: string
+  private cacheServerClient?: CacheServerClient
```
jrhender, line 53: @Harasz So is this the key part of the change? Before we had `cacheServerClient` as an asserted not-undefined and not it can be undefined?

Harasz: The key part is that we are set a reference to `cacheServerClient` after login to cache server not before. That change was just to make types more predictable.

lib/LoginStrategy.ts
```diff
@@ -105,6 +100,17 @@ export class LoginStrategy extends BaseStrategy {
...
   }
+
+  private async initCacheServer(privateKey: string, cacheServerUrl: string) {
+    const cacheServerClientInstance = new CacheServerClient({
+      privateKey,
+      provider: this.provider,
+      url: cacheServerUrl,
+    });
+    await cacheServerClientInstance.login();
+    this.cacheServerClient = cacheServerClientInstance;
```
Harasz, line 111: @jrhender The key part

jrhender: Ahhhh, I see. The key is the assignment of `this.cacheServerClient` after `login()` is completed.

lib/BaseStrategy.ts
```diff
@@ -74,7 +74,7 @@ export abstract class BaseStrategy extends Strategy {
         return this.error(err)
       }
       if (!user) {
-        return self.fail(info)
+        return this.fail(info)
```
jrhender, line 77: Why this change? Was this just a bug that you noticed?

jrhender (commented):
Another question, but I'm understanding better now

lib/LoginStrategy.ts
Harasz: Yeah, that is the point

lib/BaseStrategy.ts
Harasz: Idk, why typescript did not report that, but there is not `self` variable. While I was debugging I once run into that if statement and it throws me runtime error. It's look like someone was thinking about coding in python ;)

jrhender (approved):
Looks good to me

PR #105, opened 2021-09-22
AuthTokenVerifier: clarify isAuthenticationKey method

jrhender:


knzeng-e (approved):
This brings indeed more clarity on the purpose of the authentication field and the meaning of authentication key. Thanks @jrhender !

PR #108, opened 2021-09-23
Feat/pda 2/testing delegate auth jh cleanup

knzeng-e:
further cleanup and token checking on server response


jrhender (approved):
Looks good!

PR #114, opened 2021-09-24
feat(LoginStrategy): improve readability of ICS login

jrhender:
Previously we were awaiting cacheServerClient.login in the constructor. As this is an antipattern, changed to storing login promise as property. Then, can check promise before using cacheServerClient.

Also, removed the strategyAddress property as it was essentially equivalent to cacheServerClient.address

### DIFF
--- a/lib/LoginStrategy.ts
+++ b/lib/LoginStrategy.ts
@@ -49,8 +49,8 @@ export class LoginStrategy extends BaseStrategy {
   private readonly ipfsStore: DidStore
   private readonly acceptedRoles: Set<string>
   private readonly privateKey: string
-  private strategyAddress?: string
-  private cacheServerClient?: CacheServerClient
+  private readonly cacheServerClient?: CacheServerClient
+  private readonly cacheServerClientLoginPromise?: Promise<string>
 
   constructor(
     {
@@ -84,7 +84,12 @@ export class LoginStrategy extends BaseStrategy {
       )
     }
     if (cacheServerUrl && privateKey) {
-      this.initCacheServer(privateKey, cacheServerUrl);
+      this.cacheServerClient = new CacheServerClient({
+        privateKey,
+        provider: this.provider,
+        url: cacheServerUrl,
+      });
+      this.cacheServerClientLoginPromise = this.cacheServerClient.login();
     }
     const registrySetting = {
       abi: abi1056,
@@ -98,17 +103,6 @@ export class LoginStrategy extends BaseStrategy {
     this.acceptedRoles = acceptedRoles! && new Set(acceptedRoles)
     this.jwtSignOptions = jwtSignOptions
     this.privateKey = privateKey!
-  }
-
-  private async initCacheServer(privateKey: string, cacheServerUrl: string) {
-    const cacheServerClientInstance = new CacheServerClient({
-      privateKey,
-      provider: this.provider,
-      url: cacheServerUrl,
-    });
-    this.strategyAddress = cacheServerClientInstance.address;
-    await cacheServerClientInstance.login();
-    this.cacheServerClient = cacheServerClientInstance;
   }
 
   /**
@@ -159,7 +153,7 @@ export class LoginStrategy extends BaseStrategy {
          * Therefore, not getting userClaims
          * if address attempting to login is the address of the strategy
          */
-        this.strategyAddress === address ? [] : await this.getUserClaims(did)
+        this.cacheServerClient?.address === address ? [] : await this.getUserClaims(did)
       const verifier = new ClaimVerifier(roleClaims, this.getRoleDefinition.bind(this), this.getUserClaims.bind(this))
       const uniqueRoles = await verifier.getVerifiedRoles();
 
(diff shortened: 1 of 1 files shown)


lib/LoginStrategy.ts
```diff
@@ -49,8 +49,8 @@ export class LoginStrategy extends BaseStrategy {
   private readonly ipfsStore: DidStore
   private readonly acceptedRoles: Set<string>
   private readonly privateKey: string
-  private strategyAddress?: string
-  private cacheServerClient?: CacheServerClient
+  private readonly cacheServerClient?: CacheServerClient
+  private readonly cacheServerClientLoginPromise?: Promise<string>
```
Harasz, line 53:
```suggestion
    private isCacheServerClientAvailable: boolean
```

lib/LoginStrategy.ts
```diff
@@ -84,7 +84,12 @@ export class LoginStrategy extends BaseStrategy {
       )
     }
```
Harasz, line 85:
```suggestion
    }
    this.isCacheServerClientAvailable = false;
```

lib/LoginStrategy.ts
```diff
@@ -84,7 +84,12 @@ export class LoginStrategy extends BaseStrategy {
...
     if (cacheServerUrl && privateKey) {
-      this.initCacheServer(privateKey, cacheServerUrl);
+      this.cacheServerClient = new CacheServerClient({
+        privateKey,
+        provider: this.provider,
+        url: cacheServerUrl,
+      });
+      this.cacheServerClientLoginPromise = this.cacheServerClient.login();
```
Harasz, line 92:
```suggestion
      this.cacheServerClient.login().then(() => this.isCacheServerClientAvailable = true);
```

lib/LoginStrategy.ts
```diff
@@ -100,17 +105,6 @@ export class LoginStrategy extends BaseStrategy {
...
-      url: cacheServerUrl,
-    });
-    this.strategyAddress = cacheServerClientInstance.address;
-    await cacheServerClientInstance.login();
-    this.cacheServerClient = cacheServerClientInstance;
-  }
-
   /**
    * @description verifies issuer signature, then check that claim issued
    * no latter then `this.numberOfBlocksBack` and user has enrolled with at
```
Harasz, line 110: For line 121
```suggestion
   const didDocument = this.isCacheServerClientAvailable && this.cacheServerClient ? await this.cacheServerClient.getDidDocument(payload.iss) : await this.didResolver.read(payload.iss)
```

lib/LoginStrategy.ts
```diff
@@ -219,8 +213,17 @@ export class LoginStrategy extends BaseStrategy {
...
 
+  /**
+   * Checks the necessary initialization of cache-server has occured
+   * The pattern is described here: https://stackoverflow.com/a/49695053 
+   * @returns 
+   */
+  private async isCacheServerClientAvailable() {
+    return await this.cacheServerClientLoginPromise;
+  }
+
```
Harasz, line 224:
```suggestion

```

lib/LoginStrategy.ts
```diff
@@ -219,8 +213,17 @@ export class LoginStrategy extends BaseStrategy {
...
   async getRoleDefinition(namespace: string) : Promise<any> {
-    if (this.cacheServerClient) {
+    if (this.cacheServerClient && await this.isCacheServerClientAvailable()) {
```
Harasz, line 226:
```suggestion
    if (this.cacheServerClient && this.isCacheServerClientAvailable) {
```

lib/LoginStrategy.ts
```diff
@@ -230,7 +233,7 @@ export class LoginStrategy extends BaseStrategy {
   }
 
   async getUserClaims(did: string): Promise<Claim[]> {
-    if (this.cacheServerClient) {
+    if (this.cacheServerClient && await this.cacheServerClientLoginPromise) {
```
Harasz, line 236:
```suggestion
    if (this.cacheServerClient && this.isCacheServerClientAvailable) {
```

Harasz (commented):
@jrhender It looks a way better but it doesn't work with https://github.com/energywebfoundation/iam-cache-server/pull/174/files. I manged to fix (check my suggestions). The only thing I don't like in my suggestions is `this.cacheServerClient.login().then(() => this.isCacheServerClientAvailable = true);` We should find a better replacement for this.

jrhender:
@Harasz Thanks for the review and feedback. How does it look to you now?

Harasz (approved):
@jrhender I checked it and everything works fine. I wonder if we can find a better replacement for `.then()` in constructor.

jrhender:
@Harasz what do you not like about `.then()` in constructor? I think our options to not use `.then()` in constructor are:
1. use async function as you are currently doing on `task/ICS-62-fix-cache-server-logging` branch
2. have the promise as a property and await (as I was doing previously in this PR)

I don't like 1. because I think it is confusing to be using an `async` function without `awaiting` it. It's just quite an irregular pattern. I think using `.then()` is a much more standard approach
I think 2. is fine, it arguably makes the constructor a bit cleaner, but might be a bit more confusing than just using `.then()`

PR #133, opened 2021-09-30
test(AuthTokenVerifier): add empty doc test

jrhender:
- [x] add test for empty DID Doc
- [x] restore code which recovers eth address as DID auth verification https://github.com/energywebfoundation/passport-did-auth/blob/be6370b1a642d53f88a44493d4d2e89c3f66f815/lib/utils.ts#L74

### DIFF
--- a/lib/AuthTokenVerifier.ts
+++ b/lib/AuthTokenVerifier.ts
@@ -1,92 +1,132 @@
+import { JWT } from "@ew-did-registry/jwt";
+import { Keys } from "@ew-did-registry/keys";
+import {
+  IAuthentication,
+  IDIDDocument,
+  IPublicKey,
+  PubKeyType,
+} from "@ew-did-registry/did-resolver-interface";
+import { utils } from "ethers";
+import base64url from "base64url";
 
-import { JWT } from '@ew-did-registry/jwt';
-import { Keys } from '@ew-did-registry/keys';
-import { IAuthentication, IDIDDocument, IPublicKey } from "@ew-did-registry/did-resolver-interface";
+const { arrayify, recoverAddress, keccak256, hashMessage } = utils;
 
 export class AuthTokenVerifier {
   private _jwt = new JWT(new Keys());
   constructor(private readonly _didDocument: IDIDDocument) {}
 
-    /**
+  /**
    * @description checks a token was signed by the issuer DID or a valid authentication delegate of the issuer
    *
    * @param token
-   * @param issuerDID
    *
-   * @returns {string} issuer DID or null
+   * @returns {string} Authenticated identity DID
    */
-    public async verify(token: string, issuerDID: string): Promise<string | null> {
-        if (await this.isAuthorized(token))
-            return issuerDID
-        return null
+  public async verify(token: string): Promise<string | null> {
+    if ((await this.isIdentity(token)) || (await this.isDelegate(token))) {
+      return this._didDocument.id;
+    }
+    return null;
+  }
+
+  /**
+   * @description Determines if a token was signed with an Ethereum signature by the address in the id of the DID Document
+   * Not that JWT-compliant signatures can't be used to recover an ethereum 
+   */
+  private async isIdentity(token: string) {
+    const [encodedHeader, encodedPayload, encodedSignature] = token.split(".");
+    const msg = `0x${Buffer.from(`${encodedHeader}.${encodedPayload}`).toString(
+      "hex"
+    )}`;
+    const signature = base64url.decode(encodedSignature);
+    const hash = arrayify(keccak256(msg));
+    const claimedAddress = this._didDocument.id.split(":")[2];
+    try {
+      if (claimedAddress === recoverAddress(hash, signature)) {
+        return true;
+      }
+    } catch (_) {}
+    const digest = arrayify(hashMessage(hash));
+    try {
+      if (claimedAddress === recoverAddress(digest, signature)) {
+        return true;
+      }
+    } catch (_) {}
+    return false;
+  }
+
+  private async isDelegate(token: string) {
+    const didPubKeys = this._didDocument.publicKey;
+    if (didPubKeys.length === 0) {
+      return false;
     }
 
-    private isAuthorized = async (claimedToken: string): Promise<boolean> => {
-        //read publickey field in DID document
-        const didPubKeys = this.didDocument.publicKey
-        if (didPubKeys.length === 0) {
-            return false
+    const authenticationPubkeys = didPubKeys.filter((pubkey) => {
+      return this.isAuthenticationKey(pubkey, this._didDocument.authentication);
+    });
+    const validKeys = await this.filterValidKeys(
+      authenticationPubkeys,
+      async (pubKeyField) => {
+        try {
+          const parts = pubKeyField["publicKeyHex"]?.split("x");
+          const publickey = parts.length == 2 ? parts[1] : parts[0];
+          const decodedClaim = await this._jwt.verify(token, publickey);
+
+          return decodedClaim !== undefined;
+        } catch (error) {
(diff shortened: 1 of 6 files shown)


lib/AuthTokenVerifier.ts
```diff
@@ -1,104 +1,133 @@
...
+  public async verify(token: string): Promise<string | null> {
+    if ((await this.isIdentity(token)) || (await this.isDelegate(token))) {
+      return this._didDocument.id;
     }
+    return null;
+  }
+  
+  /**
+   * @description Authenticates by token signed with Ethereum signature
+   * Jwt can't be verified against etherum address
```
jrhender, line 33: @JGiter What does this line mean? Like the whole JWT can't be verified? Because we are using the signature from the JWT to recover the address.

test/AuthTokenVerifier.test.ts
```diff
@@ -45,56 +43,53 @@ describe("Testing AuthTokenVerifier", () => {
       sub: secondDID
     }
 
-    firstSigner = new JWT(firstKeys);
-    secondSigner = new JWT(secondKeys)
-    firstClaimToken = await firstSigner.sign(firstPayload, { algorithm: 'ES256' });
-    secondClaimToken = await secondSigner.sign(secondPayload, { algorithm: 'ES256' });
-  })
-
-  it("Checks if basic signing verification works", async () => {
```
JGiter, line 54: I've removed this test because it basically tests `Keys`

lib/AuthTokenVerifier.ts
JGiter: It authenticates whole token, but this token must be signed with ethereum signature, because verification of standard jwt requires public key

test/TestDidDocuments.ts
```diff
@@ -1,83 +1,99 @@
...
-      type: 'Secp256k1veriKey',
-      controller: '0x0E46cbecF1742DdEb29BBC56F27ad236483443b7',
-      publicKeyHex: '0x' + secondKeys.publicKey
-    }
+      id: "did:ethr:0x731ac21Aa72c1A6E15243AFBB34cA9CC073D419B#key-owner",
+      type: "Secp256k1veriKey",
+      controller: "0x731ac21Aa72c1A6E15243AFBB34cA9CC073D419B",
+      publicKeyHex:
+        "0x037cff69ef821172f5df74d3f9406679cc27aba2d96438211538deeb325c9d434d",
+    },
```
jrhender, line 36: To me it is confusing that there are two `key-owner` keys in this test and that the second one has an id and controller with a different address.
```suggestion
      id: `did:ethr:${firstWallet.address}#another-key`,
      type: "Secp256k1veriKey",
      controller: `${firstWallet.address}`,
      publicKeyHex:
        "0x037cff69ef821172f5df74d3f9406679cc27aba2d96438211538deeb325c9d434d",
    },
```

test/AuthTokenVerifier.test.ts
```diff
@@ -45,56 +43,53 @@ describe("Testing AuthTokenVerifier", () => {
...
-    const testSignature = await secondSigner.sign('Initial signed content', { algorithm: 'ES256' })
-    const secondPubKey = secondKeys.publicKey
-    const decoded = await secondSigner.verify(testSignature, secondPubKey)
-    
-    assert.strictEqual(decoded, 'Initial signed content')
+    firstClaim = await createSignWithEthersSigner(firstWallet)(firstPayload);
+    secondClaim = await createSignWithEthersSigner(secondWallet)(secondPayload);
```
jrhender, line 47: I think that it is good that we are testing Ethers signature now but I think we should also test with a JWT signer which is what I think we would get with `new JWT(keys);`. Basically, we need to test two signature approaches?

test/AuthTokenVerifier.test.ts
```diff
@@ -45,56 +43,53 @@ describe("Testing AuthTokenVerifier", () => {
...
   })
 
   it("should authenticate first DID with first Claim issuer", async () => {
```
jrhender, line 50: I know this isn't your change, but I think it would be better if the test names actually represented the use case that was being tested. More of a `why` the test is there, less of a `what`. So in this case, it would be
```suggestion
  it("should authenticate DID with created DID doc (owner pubKey) and ethers signature", async () => {
```
If you like this direction @JGiter @knzeng-e , we can try to change the other tests description.

jrhender (commented):
I like the direction it's going, just some comments

lib/AuthTokenVerifier.ts
```diff
@@ -1,104 +1,133 @@
...
-        const didPubKeys = this.didDocument.publicKey
-        if (didPubKeys.length === 0) {
-            return false
+  private async isDelegate(token: string) {
+    const didPubKeys = this._didDocument.publicKey;
+    if (didPubKeys.length === 0) {
+      // If the DID Document does not contain any publicKey, then try to login by recovering address
+      // In other words, even if the DID Document does not explictly allow an address "recovery", attempt it anyways
+      // Probably, the DID document should have a RecoverySignature public key https://identity.foundation/EcdsaSecp256k1RecoverySignature2020/
+      // In this way, it would be clear in the DID Document that the ETH address was allowed to auth as that DID
```
jrhender, line 63:
```suggestion
```

jrhender (commented):
Looks good, thanks @JGiter

test/AuthTokenVerifier.test.ts
knzeng-e: I see what you mean, I'm ok with your proposal.

test/AuthTokenVerifier.test.ts
```diff
@@ -45,56 +43,53 @@ describe("Testing AuthTokenVerifier", () => {
...
-    tokenAuthenticationVerifier = new AuthTokenVerifier(firstKeys.privateKey, firstDocument);
-    const did = await tokenAuthenticationVerifier.verify(firstClaimToken, firstPayload.iss)
+    verifier = new AuthTokenVerifier(firstDocument);
+    const did = await verifier.verify(firstClaim)
 
-    assert.strictEqual(did, issuerDID)
+    assert.strictEqual(did, firstDID)
   })
 
   it("should reject the second DID authentication with first Claim issuer", async () => {
```
knzeng-e, line 58: @jrhender is that description fits to what you have in mind ?
```suggestion
  it("should reject authentication with mismatching DID doc ", async () => {
```

jrhender: Yes, exactly. I think these kind of test descriptions are much better

test/AuthTokenVerifier.test.ts
```diff
@@ -45,56 +43,53 @@ describe("Testing AuthTokenVerifier", () => {
...
     
-    tokenAuthenticationVerifier = new AuthTokenVerifier(firstKeys.privateKey, firstDocument);
-    const did = await tokenAuthenticationVerifier.verify(secondClaimToken, firstPayload.iss)
+    verifier = new AuthTokenVerifier(firstDocument);
+    const did = await verifier.verify(secondClaim)
 
     assert.strictEqual(did, null)
   })
 
   it("should reject first DID with second Claim issuer", async () => {
```
knzeng-e, line 66: In term of `Why`-test description, maybe this one is redundant to the previous one ?

test/AuthTokenVerifier.test.ts
knzeng-e: I agree

knzeng-e (changes requested):
I took some time to review and follow the changes, thanks @JGiter, @JGiter.
A few cleanup on `AuthTokenVerifier.test` and fixing test and  we should be  all good.

knzeng-e (approved):

PR #164, opened 2021-10-13
Feat/pda 16/check issuer signature

knzeng-e:
This PR aims to verify the issuer signature and adds some tests to check the verification process.

### DIFF
--- a/test/ClaimVerifier.test.ts
+++ b/test/ClaimVerifier.test.ts
@@ -65,72 +65,49 @@ describe("ClaimVerifier", () => {
   });
 
   it('should verify Role-type claim', async () => {
-    const verifier = new ClaimVerifier(claims, getTypeRoleDefinition(secondDID), getUserClaims, getDidDocument)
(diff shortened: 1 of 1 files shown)


jrhender (approved):
I think these changes are good, so we should merge them in 😄 . But I think that they don't quite test for the issue in question 😓

jrhender:
Looks like a good start!

lib/ClaimVerifier.ts
```diff
@@ -86,5 +95,21 @@ export class ClaimVerifier {
...
+      if (claim.iss !== issuer){
+        return false
+      }
+      const claimToken = claim.issuedToken as string
+      const verifiedIssuer = await authenticationClaimVerifier.verify(claimToken)
+      if (verifiedIssuer !== issuer){
+        return false
+      }
+    })
+    return true
```
jrhender, line 112: I think this `return false` might not be quite right. Isn't it return false from the function in the map, not false for the `verifySignature` function?

jrhender (changes requested):
Potential issue, let me know what you think

PR #176, opened 2021-10-22
Fix/pda 20/fix canary version

knzeng-e:
### Summary

|             |   |
|-------------|---|
| Description |  This PR targets canary version updates so that it stays one version ahead of latest. The purpose of the PR is to keep canary versions synchronized to release branch. Hence anytime a new release is published, merging back master to develop will set canary version to the next number. The fix follows a similar pattern than https://github.com/energywebfoundation/iam-client-lib/pull/159  |
| Jira issue  |  https://energyweb.atlassian.net/browse/PDA-20  |

#### Type of request

- [x] Bug fix (non-breaking change which fixes an issue)

### List of Features

- Setting merging master into dev
- Setting Personal Access Token

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -19,10 +19,6 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     needs: cancel-previous
-    permissions: 
-      contents: write
-      deployments: write
-      packages: write
   
     steps:
       - uses: actions/checkout@v2
@@ -37,7 +33,7 @@ jobs:
         run: |
           git config --global user.email ${{ secrets.GH_EMAIL }}
           git config --global user.name ${{ secrets.GH_USER }}
-          git remote set-url origin "https://${{ github.token }}@github.com/energywebfoundation/passport-did-auth.git" > /dev/null 2>&1
+          git remote set-url origin "https://${{ secrets.GH_ACTION_TOKEN }}@github.com/energywebfoundation/passport-did-auth.git" > /dev/null 2>&1
           git reset --hard
           git fetch -u origin master:master
           git fetch -u origin develop:develop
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -61,22 +56,32 @@ jobs:
...
+          token: ${{ secrets.GH_ACTION_TOKEN }}
+
+      - name: Merge master into develop branch
         run: |
-          git checkout develop
+          git config --global user.email ${{ secrets.GH_EMAIL }}
+          git config --global user.name ${{ secrets.GH_USER }}
           git merge origin/master
-          git push origin develop
+          git push origin development
```
jrhender, line 85: I don't think this is right, I think we are using `develop` on passport-did-auth
```suggestion
          git push origin develop
```

.github/workflows/deploy.yml
```diff
@@ -61,22 +56,32 @@ jobs:
...
         uses: codfish/semantic-release-action@v1
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GH_ACTION_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
-      - name: Merge to master branch
-        if: github.ref == 'refs/heads/master'
+  sync_master_to_development_branch :
+    if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/develop'
```
jrhender, line 70: Why both `master` and `develop` here?

jrhender (changes requested):
@knzeng-e Could you please explain a bit more why this fixes the issue? I see that we will now be merging back into `develop` after a push to `master`. How will this keep the canary version ahead of "latest"? Is it because a new canary version will be pushed after the merge? Or I guess the issue is that `package.json` versions need to stay in sync?

.github/workflows/deploy.yml
knzeng-e: Oh yes indeed, my bad. Thank you @jrhender

.github/workflows/deploy.yml
knzeng-e: We want to make sure that anytime we push into develop also, we keep this synchronization with master branch.
Even though most of the time develop branch will be ahead of master, it is still possible to have some pushes on master that aren't merged into develop (for example if an admin commits directly on master)

knzeng-e:
> @knzeng-e Could you please explain a bit more why this fixes the issue? I see that we will now be merging back into `develop` after a push to `master`. How will this keep the canary version ahead of "latest"? Is it because a new canary version will be pushed after the merge? Or I guess the issue is that `package.json` versions need to stay in sync?

@jrhender I updated the description. My point is that after the merge of master into develop, if there has been a new version released on master, the `package.json` on develop will be consequently updated to start to the next version on next canary. That is how I understand it from the previous similar bug mentioned in description.

.github/workflows/deploy.yml
jrhender: So I think what you're saying is that this pipeline isn't run if an admin commits directly on `master`? I think that's the only way that commits could end up on master and not be synced back to develop?

jrhender (approved):
Looks good to me!

PR #187, opened 2021-10-28
feat(AuthTokenVerifier):  adding verification of P-256 signed token

knzeng-e:
### Summary

|             |   |
|-------------|---|
| Description |  This PR adds verification of `ECDSA P-256 SHA-256` signed tokens  |
| Jira issue  |  https://energyweb.atlassian.net/browse/PDA-22  |

#### Type of request

- [x] feature (non-breaking change)

### DIFF
--- a/lib/AuthTokenVerifier.ts
+++ b/lib/AuthTokenVerifier.ts
@@ -8,6 +8,7 @@ import {
 } from "@ew-did-registry/did-resolver-interface";
 import { utils } from "ethers";
 import base64url from "base64url";
+import * as jsonwebtoken from 'jsonwebtoken';
 
 const { arrayify, recoverAddress, keccak256, hashMessage } = utils;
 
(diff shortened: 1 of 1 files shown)


lib/AuthTokenVerifier.ts
```diff
@@ -72,7 +78,7 @@ export class AuthTokenVerifier {
           const publickey = parts.length == 2 ? parts[1] : parts[0];
           const decodedClaim = await this._jwt.verify(token, publickey);
 
-          return decodedClaim !== undefined;
+          return (decodedClaim !== undefined || this.p256verify(token, publickey));
```
JGiter, line 81: PR https://github.com/energywebfoundation/ew-did-registry/pull/349 includes verification of ES256. To enable it, `jwt.verify` should be passed array of methods [`ES256`, `ES256K`]

JGiter (changes requested):

jrhender:
@knzeng-e Thanks for your work on this PR. I think it was good to have in case Stedin wanted to test sooner. I think now that we have verification built into ew-did-registry, we should use that directly and so I think this PR can be closed.

knzeng-e:
> @knzeng-e Thanks for your work on this PR. I think it was good to have in case Stedin wanted to test sooner. I think now that we have verification built into ew-did-registry, we should use that directly and so I think this PR can be closed.

Yes @jrhender I agree. Thank you !

PR #188, opened 2021-11-01
Feat/pda 22/es256 auth using ew did

jrhender:


JGiter (approved):
Tests with secp256k1 signer would be good

PR #194, opened 2021-11-05
Feat/pda 17/code formatting tooling

knzeng-e:
### Summary

|             |   |
|-------------|---|
| Description | This PR adds some code formatting tools to enforce code readability and increase homogeneous format.|
| Jira issue  | https://energyweb.atlassian.net/browse/PDA-17  |

#### Type of request
- [x] New feature (non-breaking change which adds functionality)

### List of Features

- Prettier config is added (find [here](https://prettier.io/docs/en/options.html) the different options that can be set for prettier)
- Eslint config is updated to integrate prettier checks
- Prettier and eslint are run as a precommit-hook using husky

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-module-utils": "^2.7.0",
     "eslint-plugin-import": "^2.25.1",
+    "eslint-plugin-prettier": "^4.0.0",
     "express": "^4.17.1",
     "husky": "^7.0.2",
     "iam-client-lib": "^3.1.0",
@@ -91,6 +92,7 @@
     "mocha": "^9.1.1",
     "morgan": "^1.10.0",
     "passport-jwt": "^4.0.0",
+    "prettier": "^2.4.1",
     "run-with-testrpc": "^0.3.1",
     "semantic-release": "^18.0.0",
     "supertest": "^6.1.6",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "start-rpc": "run-with-testrpc -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" --port 8544 --chainId 1337 --accounts 20 --gasLimit=10000000",
     "ganache": "ganache-cli -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" --port 8544 --accounts 20 --networkId=9 --gasLimit=10000000",
-    "prepare": "npm run build",
+    "prepare": "npm run build && husky install",
```
jrhender, line 13: Why `husky install` on `prepare`? It seems `prepare` is run before `npm install` and `npm publish`?

JGiter (approved):

package.json
knzeng-e: I saw in [documentation](https://typicode.github.io/husky/#/) and on the [github page](https://github.com/typicode/husky#usage) that setting it on the `prepare script` enables git hooks directly after the install and long before `npm publish`

knzeng-e: The consequence of that is that at each commit, the eslint & prettier checks will be fired, and the commit wont be possible if there is any error.

JGiter: It will create husky.sh in .husky/_/

jrhender: Thanks!

jrhender (approved):

PR #216, opened 2021-12-02
bug/PDA-23 not issued token in claim

JGiter:
`issuedToken` missing in claims read from document. There are also other places in code which potentially could lead to similar bugs. They are all caused by optional properties of `Claim` https://github.com/energywebfoundation/passport-did-auth/blob/66b4a242d64a41eda31dfa20f53d5240b95c2af5/lib/LoginStrategy.types.ts#L26. I've set `claimType` and `claimTypeVersion` required as it done in cache-server and iam-lib

- `issuedToken`, `claimType` and `claimTypeVersion` set to required
- `ClaimVerifier.verifySignature` is changed to verify only issuer signature on claim (https://energyweb.atlassian.net/browse/PDA-23?focusedCommentId=18110)


lib/utils.ts
```diff
@@ -69,3 +70,14 @@ export function lookup(
   }
   return null;
 }
+
+export function isOffchainClaim(claim: any): claim is OffchainClaim {
```
jrhender, line 74: I think this is the core of the fix, right?

jrhender (approved):
Looks good, thanks @JGiter

jrhender (changes requested):
Some conflicts to resolve now

jrhender (approved):
Looks good, thanks @JGiter

PR #217, opened 2021-12-02
chore: update prettier config

jrhender:
In this way, it can be applied to the whole project
(as correct files are ignored)


.prettierrc.json
```diff
@@ -1,5 +1,5 @@
 {
-    "tabSize": 2,
```
jrhender, line 2: I think `tabWidth` is the correct option: https://prettier.io/docs/en/options.html#tab-width

JGiter (approved):

PR #221, opened 2021-12-04
fix(cacheServerClient): destructure axios response in login

jrhender:
Not sure how it worked before.
But `data` is a property of AxiosResponse type which is returned from `post`. `token` isn't a direct property of the response.
This how it is in `iam-client-lib` as well: https://github.com/energywebfoundation/iam-client-lib/blob/788f863d99b73599983567e265a8a911e3f1261a/src/modules/cacheClient/cacheClient.service.ts#L64


jrhender:
@Harasz @JGiter I merged this because I think it is pretty straightforward but please take a look when you have time in case I've missed something obvious 😄

PR #229, opened 2021-12-08
feat(logs): hide sensitive data from logs

Passerino:
introduced Logger responsible for redacting sensitive data


knzeng-e (approved):
It looks good to me @Passerino, thanks !
I requested @jrhender's review just in case I'm missing anything.

PR #265, opened 2022-01-24
rename push token

JGiter:


jrhender (approved):
I think this is fine, but why the name `PUBLISH_TOKEN` over `GH_ACTION_TOKEN`?

JGiter:
> I think this is fine, but why the name `PUBLISH_TOKEN` over `GH_ACTION_TOKEN`?

I renamed `PUBLISH_TOKEN` too on `GH_PUSH_TOKEN`. This is mostly to avoid confusion with github token

jrhender:
Nice, `GH_PUSH_TOKEN` makes sense to me 👍

PR #269, opened 2022-01-24
clean credentials after checkout

JGiter:


.github/workflows/deploy.yml
```diff
@@ -32,7 +33,7 @@ jobs:
         run: |
           git config --global user.email ${{ secrets.GH_EMAIL }}
           git config --global user.name ${{ secrets.GH_USER }}
-          git remote set-url origin "https://${{ secrets.GH_ACTION_TOKEN }}@github.com/energywebfoundation/passport-did-auth.git" > /dev/null 2>&1
+          git remote set-url origin "https://${{ secrets.GITHUB_TOKEN }}@github.com/energywebfoundation/passport-did-auth.git" > /dev/null 2>&1
```
jrhender, line 36: I wonder if this is the key change!

jrhender (approved):

.github/workflows/deploy.yml
JGiter: No, this step was passed in CI. The error happened latter - in `Deploy on develop branch`. I am also thinking of removing `Configure Git` step since we are using actions/checkout now

PR #273, opened 2022-01-25
improve publishing

JGiter:
- configuring `git` assigned to gh action
- publish on `master` and `develop` integrated in one step

Synchronization with development is still separate job, because it is consists from several steps. Otherwise each step would include `if`


jrhender (approved):
Looks good to me!

PR #282, opened 2022-02-14
feat: add support for old DID format

Harasz:


jrhender (commented):
Looks great. What do you think of using these helper methods from ew-did-registry for parsing the DID? https://github.com/energywebfoundation/ew-did-registry/blob/6799fe007c1dc9a1e0b43f973fb516938e545673/packages/did/src/utils/parser.ts#L26
Would they help?

JGiter (approved):

Harasz:
@jrhender I could you use it, but i don't think it will simplify our logic in `didUnification`. I would rather add new function `@ew-did-registry/did` that will append a chain if not presented in did :thinking:

lib/LoginStrategy.ts
```diff
@@ -248,18 +249,56 @@ export class LoginStrategy extends BaseStrategy {
...
+  }
+
+  /**
+   * @param {string} did
+   * @returns {string} DID address in format "did:" method-name ":" method-specific-id ":" address
+   */
+  didUnification(did: string): string {
+    const didParts = did.split(':');
+
+    if (didParts.length === 4) return did;
```
jrhender, line 293: Here's what I'm thinking for using ew-did-registry. I think this would set us up to basically move the `didUnification` function into ew-did-registry/did directly. What do you think?
```suggestion
    const { foundChainInfo } = getDIDChain(did)
    if (foundChainInfo) return did;
    
    const didParts = did.split(':');
```

Harasz: Yeah, I was thinking on moving `didUnification` to `@ew-did-registry/did`. The suggestion you proposed just replace one simple line of code, could you explain me why this is a better approach?

jrhender: I agree it's not a big difference however, I think it is better because it reused the function of `getDIDChain`, which answers the question of whether or not the chainID needs to be added. I know it is a small thing but if it ever does get more complex, we can put the additional complicity in the provided method.
From a non-functional standpoint, it also makes it easier to search in the future for all of the places where we have used `getDIDChain` to have a better idea of what might break if we introduce a breaking change, for example.
Do you agree with this or not really?

Harasz: Yeah, you got me. I will update it

jrhender (approved):
Looks good!

PR #284, opened 2022-02-17
fix: omit invalid claims in off-chain claims

Harasz:
Malicious claim is a claim that have a role namespace instead of
DID address because of the bug we have in the past.

### DIFF
--- a/lib/LoginStrategy.ts
+++ b/lib/LoginStrategy.ts
@@ -17,7 +17,7 @@ import {
 } from './LoginStrategy.types';
 import { PublicResolver__factory } from '../ethers/factories/PublicResolver__factory';
 import { PublicResolver } from '../ethers/PublicResolver';
-import { Methods, getDIDChain } from '@ew-did-registry/did';
+import { Methods, getDIDChain, isValidErc1056 } from '@ew-did-registry/did';
 import { DidStore } from '@ew-did-registry/did-ipfs-store';
 import { CacheServerClient } from './cacheServerClient';
 import { ClaimVerifier } from './ClaimVerifier';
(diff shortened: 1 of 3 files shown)


lib/LoginStrategy.ts
```diff
@@ -251,23 +251,41 @@ export class LoginStrategy extends BaseStrategy {
...
 
-    const transformClaim = (claim: OffchainClaim): OffchainClaim => {
+    const transformClaim = (
+      claim: OffchainClaim
+    ): OffchainClaim | undefined => {
       const transformedClaim = { ...claim };
-      const didFormatFields = ['iss', 'sub', 'subject', 'did'];
+      const didFormatFields = ['iss', 'sub', 'subject', 'did', 'signer'];
+
+      let isMaliciousClaim = false;
```
jrhender, line 260: I think I would be more specific with the error. I don't think it is necessarily malicious.
```suggestion
      let invalidDIDProperty = false;
```

lib/LoginStrategy.ts
```diff
@@ -251,23 +251,41 @@ export class LoginStrategy extends BaseStrategy {
...
-          transformedClaim[key] = this.didUnification(transformedClaim[key]);
+          try {
+            transformedClaim[key] = this.didUnification(transformedClaim[key]);
+          } catch (err) {
+            if (err instanceof Error && err.message === 'Invalid DID') {
+              isMaliciousClaim = true;
+              return;
+            }
+            throw err;
+          }
```
jrhender, line 272: Rather than catching the exception, I think it would be better if we test the case first. We can use the helper function from `ew-did-registry`
```suggestion
            const expectedEthrDID = transformedClaim[key];
            if (isValidErc1056(expectedEthrDID)) {
              transformedClaim[key] = this.didUnification(transformedClaim[key]);
            }
            else {
              isMaliciousClaim = true;
            }
            return;
```

jrhender (changes requested):
Thanks for the PR. I added a suggestion for improvement

lib/LoginStrategy.ts
```diff
@@ -251,23 +251,38 @@ export class LoginStrategy extends BaseStrategy {
...
       const transformedClaim = { ...claim };
-      const didFormatFields = ['iss', 'sub', 'subject', 'did'];
+      const didFormatFields = ['iss', 'sub', 'subject', 'did', 'signer'];
+
+      let invalidDIDProperty = false;
 
       Object.keys(transformedClaim).forEach((key) => {
         if (didFormatFields.includes(key)) {
-          transformedClaim[key] = this.didUnification(transformedClaim[key]);
+          const expectedEthrDID = transformedClaim[key];
```
JGiter, line 264: What this variable is for? Is there some reason why `transformedClaim[key]` can't be passed to `isValidErc1056`?

JGiter (approved):

lib/LoginStrategy.ts
Harasz: Just to make code more descriptive about what `transformedClaim[key]` should be

JGiter (approved):

jrhender (approved):

PR #291, opened 2022-03-11
bug/IS-16 approved claim not verified

JGiter:
https://energyweb.atlassian.net/browse/IS-16


lib/LoginStrategy.ts
```diff
@@ -75,10 +90,22 @@ export class LoginStrategy extends BaseStrategy {
   ) {
     super(options);
     this.claimField = claimField;
-    this.provider = new providers.JsonRpcProvider(rpcUrl);
-    this.ensResolver = PublicResolver__factory.connect(
-      ensResolverAddress,
-      this.provider
+    this.provider = new JsonRpcProvider(rpcUrl);
+    const { chainId, didRegistryAddress, ensRegistryAddress } =
+      chainConfigs()[rpcUrl];
```
jrhender, line 95: Instead of hardcoding the `chainId` based on `rpcUrl` could we instead get the `chainId` from the provider?

JGiter: Yes, but it will require initialization after construction because getting chain id is asynchronous. Can we allow this change?

jrhender: Hmm, how about this approach:
- Regardless of Volta/EWC, we add all of the Volta/EWC resolvers that we are aware of.
- We optionally allow specification of new resolvers in the constructor (as you have done), and also require the `chainId` for each resolver

JGiter: @jrhender Though we still need chain id to get ens and did registry addresses

JGiter: I am not sure. This way we introduce possibility for mismatch chain id and rpcUrl. It is more consistent to provide ens and DID registries along with resolvers

jrhender: > more consistent to provide ens and DID registries along with resolvers
Ok, this makes sense to me, good idea 👍

jrhender (approved):
Looks great, thanks @JGiter !

PR #293, opened 2022-03-15
bug/PDA-45 duplicated verified roles

JGiter:


jrhender:
@JGiter thanks for the PR. Is this related to a bug in Jira? Or you found it some other way?

Harasz (approved):

PR #297, opened 2022-04-20
feat: add token payload validation

### LINKED ISSUE #294: Process crashes when malformed identity token provided.
After calling an endpoint covered with LoginStrategy with malformed identity token like for instance
```json
{
    "identityToken": "sdfsdfsdf"
}
```
the whole nodejs process crashes after an exception is thrown in the following line: https://github.com/energywebfoundation/passport-did-auth/blob/1df78a155f12c09429d5652de3f61668850142c2/lib/LoginStrategy.ts#L146-L146

```
/mydid-protected-service/node_modules/passport-did-auth/dist/lib/LoginStrategy.js:110
            const iss = this.didUnification(payload.iss);
                                                    ^
TypeError: Cannot read properties of null (reading 'iss')
```

This makes it very easy to perform DOS attacks.

Harasz:


lib/LoginStrategy.ts
```diff
@@ -156,11 +161,12 @@ export class LoginStrategy extends BaseStrategy {
     const userAddress = addressOf(userDid);
 
     try {
+      const claimBlockNumber =
```
JGiter, line 164: we can handle any type of claimData.blockNumber by
```
if (
   ...
latestBlock.sub(this.numberOfBlocksBack).ge(payload.claimData.blockNumber)
```

lib/LoginStrategy.ts
```diff
@@ -334,4 +340,39 @@ export class LoginStrategy extends BaseStrategy {
     }
     return `${didParts[0]}:${didParts[1]}:${chainName}:${didParts[2]}`;
   }
+
+  validateTokenPayload(payload: unknown): payload is ITokenPayload {
```
JGiter, line 344:
```suggestion
  isTokenPayload(payload: unknown): payload is ITokenPayload {
```

Harasz: Could you explain why this is a better name?

JGiter: It is unclear what validation mean

JGiter (approved):

PR #298, opened 2022-04-22
docs(README): add section about token payload structure

### LINKED ISSUE #46: Missing documentation on the login payload
When using `passport-did-auth` with your own web application it's pretty much unknown what is expected to be sent to your `login` endpoint which uses strategies from `passport-did-auth`. This is especially important as use of `iam-client-lib` is not mandatory and users can construct login data `identityToken` manually.

Harasz:


README.md
```diff
@@ -8,6 +8,21 @@
...
+```
+{
+  claimData: {
+    blockNumber: number;
+  };
+  iss: string;
+}
+```
+
+where the `iss` is DID od the subject and `blockNumber` is block number (or height) of the most recently mined block.
```
JGiter, line 24:
```suggestion
where the `iss` is DID of the subject and `blockNumber` is block number (or height) of the most recently mined block.
```

JGiter (approved):

lib/LoginStrategy.types.ts
```diff
@@ -21,7 +21,7 @@ export interface DecodedToken {
   sub: string;
 }
 
-export interface ITokenPayload extends DecodedToken {
+export interface ITokenPayload {
```
jrhender, line 24: Can we remove the `DecodedToken` interface now? Or is it still used?

Harasz: No it's still used as a type for decoded `claim.issuedToken`

PR #304, opened 2022-07-06
fix: fetching off-chain claims from did document

Harasz:


lib/LoginStrategy.types.ts
```diff
@@ -1,18 +1,19 @@
...
   namespace: string;
   owner: string;
-  definition: IRoleDefinition;
+  definition: IRoleDefinitionV2;
 }
 
 export interface OffchainClaim {
   claimType: string;
   claimTypeVersion: number;
-  issuedToken: string;
```
JGiter, line 14: Why is this?

Harasz: There is now `issuedToken` in claim. All data is already decoded from ipfs and token, and we never add this token to decoded data

lib/ClaimVerifier.ts
```diff
@@ -1,8 +1,6 @@
-import { OffchainClaim, DecodedToken } from './LoginStrategy.types';
-import * as jwt from 'jsonwebtoken';
+import { OffchainClaim } from './LoginStrategy.types';
 import { IRoleDefinition } from '@energyweb/credential-governance';
 import { IDIDDocument } from '@ew-did-registry/did-resolver-interface';
-import { ProofVerifier } from '@ew-did-registry/claims';
 
 export class ClaimVerifier {
```
JGiter, line 5: Even if this is not relates to this PR I wonder can ClaimVerifier be replaced by CredentialResolver and ClaimIssuerVerification from vc-verification now?

JGiter (approved):
Probably it is has to be done in separate task

lib/ClaimVerifier.ts
Harasz: I think so, I will create a ticket

PR #307, opened 2022-08-01
feat: verify hierarchy of issuers for claims

nichonien:
This PR makes changes to incorporate verification of all issuers in the hierarchy for a claim.


jrhender (commented):
I think this is going well, left some comments.

One point is that `passport-did-auth` is currently relying on the ssi-hub for DID document and role resolution. I think if we rely on the resolver classes directly from `@energyweb/vc-verification`. So we may need to write our own classes in `passport-did-auth` that leverage ssi-hub

lib/LoginStrategy.ts
```diff
@@ -390,30 +340,4 @@ export class LoginStrategy extends BaseStrategy {
...
-
-      if (hash instanceof Uint8Array) {
-        return Boolean(CID.decode(hash));
-      }
-
-      return Boolean(CID.asCID(hash));
-    } catch (e) {
-      return false;
-    }
-  }
```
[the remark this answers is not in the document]
nichonien: @jrhender We are not checking as of now. Would move this to `ew-credentials` and add use for validation.

package.json
```diff
@@ -8,7 +8,7 @@
     "/dist"
   ],
   "scripts": {
-    "start-rpc": "run-with-testrpc -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" --port 8544 --chainId 1337 --accounts 20 --gasLimit=10000000",
+    "start-rpc": "run-with-testrpc -m \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" --port 8544 --chainId 73799 --accounts 20 --gasLimit=10000000",
```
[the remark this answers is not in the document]
nichonien: Yeah, while fetching `roleDefinition` the `DomainReader`checks if the chainID is either of the `Volta` or `etc` and if not then it throws `No did:ethr networkName found for chainID`.

lib/LoginStrategy.ts
```diff
@@ -75,6 +82,11 @@ export class LoginStrategy extends BaseStrategy {
       acceptedRoles,
       ...options
     }: LoginStrategyOptions,
+    private issuerResolver: IssuerResolver,
+    private revokerResolver: RevokerResolver,
+    private revocationVerification: RevocationVerification,
+    private credentialResolver: CredentialResolver,
+    private verifyProof: (vc: string, proof_options: string) => Promise<any>,
```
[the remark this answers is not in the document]
nichonien: I will look in it.

lib/RoleCredentialResolver.ts
```diff
@@ -0,0 +1,233 @@
...
+    cacheServerUrl: string
+  ) {
+    this._ipfsStore = didStore;
+    this._resolver = new Resolver(provider, registrySetting);
+    this._cacheServerClient = new CacheServerClient({
+      privateKey,
+      provider,
+      url: cacheServerUrl,
+    });
+    this._cacheServerClient.login();
```
jrhender: I think there is quite a bit of duplicate from `IpfsCredentialResolver` (https://github.com/energywebfoundation/ew-credentials/blob/develop/packages/vc-verification/src/Ipfs-credential-resolver.ts) here. 

I think we should either accept an IpfsCredentialResolver instance in the constructor and then delegate calls to the IpfsResolver if the cacheServerClient isn't available or construct an instance using the params.

lib/RoleIssuerResolver.ts
```diff
@@ -0,0 +1,61 @@
...
+  private _domainReader: DomainReader;
+  private _cacheServerClient: CacheServerClient;
+
+  constructor(
+    domainReader: DomainReader,
+    provider: providers.Provider,
+    privateKey: string,
+    cacheServerUrl: string
+  ) {
+    this._domainReader = domainReader;
```
jrhender, line 22: Similar to CredentialResolver, I think we use an instance of `EthersProviderIssuerResolver ` to reduce the duplication of this code

lib/ClaimVerifier.ts
```diff
@@ -33,44 +29,38 @@ export class ClaimVerifier {
...
 
-    const role = await this.getRoleDefinition(claim.claimType);
-    if (!role) {
-      return null;
-    }
-
-    if (role.version !== claim.claimTypeVersion) {
-      return null;
-    }
+    if (credential.payload) {
```
jrhender, line 45: Personally, I think the code is easier to read if the `if` statement is inverted and we handle the negative condition immediately instead of needing to scan down to see what happens if this isn't true
```suggestion
    if (!credential?.payload) {
      // Maybe add some log here
      return null;
    }
```

lib/RoleCredentialResolver.ts
```diff
@@ -0,0 +1,164 @@
...
+   * Fetches all the Verifiable Credential belonging to the subject DID
+   * @param did subject DID
+   * @returns VerifiableCredential<RoleCredentialSubject> list
+   */
+  async credentialsOf(
+    did: string
+  ): Promise<VerifiableCredential<RoleCredentialSubject>[]> {
+    if (this._cacheServerClient?.isAvailable) {
+      const vcs =
+        await this._cacheServerClient.getVerifiableCredentialBySubject(did, {
```
jrhender, line 157: Thinking about it a bit more, I think this might be a problem as the identity of running passport-did-auth will likely not be able to access all of the credentials of the issuer due to authorization protection on ssi-hub. On ssi-hub, only a valid issuer or the subject can read the subject's credentials, I believe.

jrhender (commented):
I think it's looking good, but we might have an auth to ssi-hub issue

lib/ClaimVerifier.ts
```diff
@@ -33,44 +30,46 @@ export class ClaimVerifier {
...
     //   return null;
     // }
 
-    const role = await this.getRoleDefinition(claim.claimType);
-    if (!role) {
-      return null;
-    }
-
-    if (role.version !== claim.claimTypeVersion) {
+    if (!credential?.payload) {
```
JGiter, line 46: `credential isn't optional, so we don't need to null-check it
```suggestion
    if (credential?.payload) {
```

JGiter (changes requested):

lib/ClaimVerifier.ts
nichonien: Aah, right!

lib/RoleCredentialResolver.ts
```diff
@@ -0,0 +1,189 @@
...
+import {
+  IpfsCredentialResolver,
+  isEIP191Jwt,
+  isVerifiableCredential,
+  transformClaim,
+  filterOutMaliciousClaims,
+} from '@energyweb/vc-verification';
+import * as jwt from 'jsonwebtoken';
+
+export class RoleCredentialResolver implements CredentialResolver {
```
JGiter, line 21: Is it make sense to implement CredentialResolver for resolution only from ssi-hub. Say `SsiHubCredentialResolver`

lib/ClaimVerifier.ts
```diff
@@ -33,44 +30,46 @@ export class ClaimVerifier {
...
-    ) {
-      return {
-        name: role.roleName,
-        namespace: claim.claimType,
-      };
-    } else if (role.issuer?.issuerType === 'Role' && role.issuer.roleName) {
-      const issuerClaims = await this.getOffchainClaims(claim.iss);
-      const issuerRoles = issuerClaims.map((claim) => claim.claimType);
-      if (issuerRoles.includes(role.issuer.roleName)) {
+      if (role.version !== credential.payload.claimData.claimTypeVersion) {
```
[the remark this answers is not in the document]
nichonien: I think this is the right place for the check, even if we move this to `getRoleDefinition` then also it will be a call inside it to validate and would run every time `getRoleDefinition` is called and also it just relies on namespace this way.

lib/RoleCredentialResolver.ts
nichonien: IMO this would extra overhead because if there is no `sis-hub` available then we would have to go for `IpfsCredentialResolver`. So there is extra validation and more classes for same purpose.

lib/ClaimVerifier.ts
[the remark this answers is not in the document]
JGiter: The purpose of moving this check in `getRoleDefinition` is to avoid repeated checks

lib/RoleCredentialResolver.ts
JGiter: > So there is extra validation

Do you mean null-check of ssi-hub?

 > more classes for same purpose

CredentialResolver is interface, it is supposed to have different implementations

lib/ClaimVerifier.ts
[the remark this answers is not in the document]
nichonien: I'm not really following. You call `getRoleDefinition` once for one credential, so even if we move this validation inside of `getRoleDefinition` then also it would be performed once for that credential. How this will avoid repeated calls? In both case this validation will happen the same number of times the `getRoleDefinition` is called. Am I missing something?

lib/cacheServerClient.ts
```diff
@@ -166,13 +179,58 @@ export class CacheServerClient {
...
+          namespace,
+        },
+      }
+    );
+    let vcs;
+    data.map((credential) => {
+      vcs.push(credential.vp.verifiableCredential);
+    });
+    return vcs;
+  }
```
jrhender, line 230: Can this code be removed as I don't think we're using the `claims` interface anymore?

lib/LoginStrategy.ts
```diff
@@ -75,6 +82,13 @@ export class LoginStrategy extends BaseStrategy {
       acceptedRoles,
       ...options
     }: LoginStrategyOptions,
+    private issuerResolver: IssuerResolver,
+    private revokerResolver: RevokerResolver,
+    private credentialResolver: CredentialResolver,
+    private verifyProof: (
```
jrhender, line 88: It seems like we don't need to hold on to references to these classes beyond the constructor
```suggestion
    issuerResolver: IssuerResolver,
    revokerResolver: RevokerResolver,
    private credentialResolver: CredentialResolver,
    verifyProof: (
```

lib/RoleCredentialResolver.ts
nichonien: We had `IpfsCredentialResolver` for getting credential from IPFS, but then we wanted to get DID Doc from `sis-hub` so we implemented `RoleCredentialResolver` which now has both ability (either to get DID Doc from blockchain of sis-hub). So if we create `SsiHubCredentialResolver` then also we would have a third class which either gets credential through `SsiHubCredentialResolver` or `IpfsCredentialResolver` keeping a validation that if `sis-hub` is available or not. So I am not actually understanding how this new `SsiHubCredentialResolver` would be efficient?

lib/LoginStrategy.ts
```diff
@@ -255,84 +282,21 @@ export class LoginStrategy extends BaseStrategy {
...
-        .filter(isOffchainClaim)
-        .map(transformClaim)
-        .filter(filterOutMaliciousClaims);
-    }
-    const didDocument = await this.didResolver.read(did);
-    const services = didDocument.service || [];
-    return (
-      await Promise.all(
-        services.map(async ({ serviceEndpoint, ...rest }) => {
-          if (!this.isCID(serviceEndpoint)) {
```
jrhender, line 317: Do we need this `isCID` check in the `RoleCredentialResolver`?

jrhender (commented):
Nearly there I think. Added some additional comments though

lib/LoginStrategy.ts
```diff
@@ -75,6 +82,13 @@ export class LoginStrategy extends BaseStrategy {
       acceptedRoles,
       ...options
     }: LoginStrategyOptions,
+    private issuerResolver: IssuerResolver,
+    private revokerResolver: RevokerResolver,
+    private credentialResolver: CredentialResolver,
+    private verifyProof: (
+      vc: string,
+      proof_options: string
+    ) => Promise<unknown>,
```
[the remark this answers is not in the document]
nichonien: `Unknown` is replacement for `any`, since `eslint` does not allow `any` and gives a warning.

lib/LoginStrategy.ts
nichonien: Hmm, right. I will add the validation in `RoleCredentialResolver`

lib/LoginStrategy.ts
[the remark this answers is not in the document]
jrhender: Right, doesn't `verifyProof` need to return a result though? If so, can we describe the type of that result?

Or is the expectation that it throws an exception if verification fails? If that's the case, I would expect the return type to be `Promise<void>`.

Perhaps we could add constructor documentation to clarify this?

lib/ClaimVerifier.ts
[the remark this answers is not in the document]
nichonien: `getRoleDefinition` is a general purpose function which fetches the roleDefinition based on namespace. So it may or mayn't have use case to validate the roleDefinition against the `claimTypeVersion`. So IMO it is not a mandatory check.

lib/LoginStrategy.ts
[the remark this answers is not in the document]
nichonien: It would return a result, but then that result would be based on the implementation of this callBack method. So for ex, `verifyCredential` from `didkit-wasm-node` would return differently then some other method from other ssi-library and the same behaviour would be for throwing errors. This is why it is unknown. 

I would add some documentation.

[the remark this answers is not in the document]
jrhender: Hmm, I don't think the interface is very valuable if it can't be trusted at all. I think we should require that implementations conform to an interface. So if one wants to use `didkit-wasm-node` or a different implementation, one should need to write an adapter around it so that it conforms to the interface, IMO

lib/RoleIssuerResolver.ts
```diff
@@ -0,0 +1,65 @@
...
+  DomainReader,
+  IIssuerDefinition,
+} from '@energyweb/credential-governance';
+import { IssuerResolver } from '@energyweb/vc-verification';
+import { providers, utils } from 'ethers';
+import { CacheServerClient } from './cacheServerClient';
+import { EthersProviderIssuerResolver } from '@energyweb/vc-verification';
+
+/**
+ * Resolves issuers definition by either by reading smart contract state via an Ethers provider or from ssi-hub (cache server)
```
JGiter, line 11: IMO it would be clearer this way
```suggestion
 * Resolves issuer definition from ssi-hub, if it is specified. Otherwise read issuer from blockchain
```

JGiter (commented):
RoleIssuerResolver and RoleRevokerResolver has a lot in common. Is it possible to combine them in RoleResolver?

nichonien:
> RoleIssuerResolver and RoleRevokerResolver has a lot in common. Is it possible to combine them in RoleResolver?

@JGiter It was a single class initially in `ew-credentials` but then after some discussion we spliced it into two. Here is the initial comment https://github.com/energywebfoundation/ew-credentials/pull/27#discussion_r891328624

JGiter (approved):

PR #318, opened 2022-08-18
fix: fixing login strategy crashing and adding infura credentials

artursudnik:
This PR:
- fixes an issue when an entire process crashes when SSI Hub `/DID/{did}` endpoint responds with `500 Interal Server Error` code
- adds credentials to the ipfs requests


jrhender (approved):
Looks good. The one concern that I have is that `package-lock.json` and `package.json` seem out of sync. I noticed the `version` prop is different between the files. I think this is what caused the lint GHA to fail? https://github.com/energywebfoundation/passport-did-auth/runs/7902746713?check_suite_focus=true
Maybe take a quick look to see if we should fix this before merging `master`. If you think it's alright feel free to merge and see what happens, IMO.

nichonien (approved):

artursudnik:
Oh, I was wondering why `npm ci` works on my local copy and not on GHA.

JGiter (approved):

artursudnik:
GHA also fails after syncing versions between `package.json` and `package-lock.json`. I am trying to understand what is going on before merging.

artursudnik:
Looks like GHA for this PR was executed with nodejs v18 for some reason and this is why it crashed when executing `npm ci`. It is fixed now.

PR #326, opened 2022-09-02
fix: check for expiration date

whitneypurdum:
- check for expiration date in Claim Verifier


lib/ClaimVerifier.ts
```diff
@@ -72,4 +83,41 @@ export class ClaimVerifier {
...
+      return false;
+    }
+    try {
+      if (rolePayload.credentialStatus) {
+        await this.statusListEntryVerificaiton.verifyCredentialStatus(
+          rolePayload?.credentialStatus as StatusList2021Entry
+        );
+      }
+    } catch (error) {
+      const credential =
```
artursudnik, line 108: Should this logic be executed for any kind of error?

nichonien: There are three different errors thrown actually, for a insecure status list url ( not a possible case for credential issued with IAM stack), error for invalid statusListCredential (not a possible case for EWF Role) and then if the credential is revoked  which is handled here. 

Do you think if we should handle all the error scenarios?

artursudnik: What I mean is that possibly not only these errors can be thrown. For instance, when `this.statusListEntryVerificaiton.verifyCredentialStatus` causes requests to be made to an external system, network issues can also happen.

artursudnik: I see that for example network request is made [here](https://github.com/energywebfoundation/passport-did-auth/blob/22340f45c29f37ffa0c5d3c762aa5100c79817b9/node_modules/@ew-did-registry/revocation/src/implementations/status-list-entry-verification.ts#L95-L95)

nichonien: Make sense, I should probably catch only the error which is expected then.

JGiter (approved):

lib/ClaimVerifier.ts
artursudnik: Or rethrow the error if it does not fit in a possible scenario you have described.

PR #329, opened 2022-09-13
feat: pass cacheserverclient instance to resolvers

nichonien:
This PR makes changes to use `cacheServerInstance` instead of initialising it in resolvers classes.


nichonien:
Closing this PR, since this is not a feasible solution.

PR #330, opened 2022-09-13
fix: refactor role def resolution

nichonien:
The PR fixes error occurred due to undefined roleDefinition coming from SSI-Hub. Also, it passes the actual namespace instead of namehash for fetching roleDefinition form SSI-Hub

### DIFF
--- a/lib/RoleIssuerResolver.ts
+++ b/lib/RoleIssuerResolver.ts
@@ -3,7 +3,7 @@ import {
   IIssuerDefinition,
 } from '@energyweb/credential-governance';
 import { IssuerResolver } from '@energyweb/vc-verification';
-import { providers, utils } from 'ethers';
+import { providers } from 'ethers';
 import { CacheServerClient } from './cacheServerClient';
 import { EthersProviderIssuerResolver } from '@energyweb/vc-verification';
 import { Logger } from './Logger';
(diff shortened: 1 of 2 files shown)


lib/RoleRevokerResolver.ts
```diff
@@ -58,11 +58,10 @@ export class RoleRevokerResolver implements RevokerResolver {
...
       );
-      return (
-        await this._cacheServerClient.getRoleDefinition({
-          namespace: resolvedNamespace,
-        })
-      ).revoker;
+      const roleDef = await this._cacheServerClient.getRoleDefinition({
+        namespace: resolvedNamespace,
+      });
+      return roleDef.issuer;
```
jrhender, line 64: @nichonien should it be `revoker` here?
```suggestion
      return roleDef.revoker;
```

jrhender (commented):
@nichonien it looks the same to me, just with an intermediary variable. Could you explain the bug a bit more please?

nichonien:
> @nichonien it looks the same to me, just with an intermediary variable. Could you explain the bug a bit more please?

Yes, I too thought this would work but it throws error : 
Screenshot 2022-09-13 at 4 47 49 PM

nichonien:
Not actually, the issue still remains after this change as well

nichonien:
@jrhender The issue was with the namespace being passed to `cacheServerClient` and also there was no validation to check if `roleDefinition` is undefined. I made the changes and this should fix our problems.

lib/RoleIssuerResolver.ts
```diff
@@ -48,22 +48,19 @@ export class RoleIssuerResolver implements IssuerResolver {
...
-      Logger.info(
-        `IIssuerDefinition for namespace: ${namespace} fetched from SSI-Hub`
-      );
-      return (
-        await this._cacheServerClient.getRoleDefinition({
-          namespace: resolvedNamespace,
-        })
-      ).issuer;
+      const roleDef = await this._cacheServerClient.getRoleDefinition({
+        namespace: namespace,
```
JGiter, line 53:
```suggestion
        namespace,
```

JGiter (approved):

jrhender:
> @jrhender The issue was with the namespace being passed to `cacheServerClient` and also there was no validation to check if `roleDefinition` is undefined. I made the changes and this should fix our problems.

Ok, thanks @nichonien

PR #331, opened 2022-09-14
fix :filter claims before verification

nichonien:


lib/LoginStrategy.ts
```diff
@@ -266,8 +266,11 @@ export class LoginStrategy extends BaseStrategy {
         this.cacheServerClient?.address === userAddress
           ? []
           : await this.credentialResolver.eip191JwtsOf(userDid);
+      const claims = userClaims.filter((claim) =>
```
jrhender, line 269: I think the variable name can be improved here. `claims` is less descriptive than `userClaims`, so it unintuitive that is the result of a `filter`, IMO.
```suggestion
      const roleClaims = userClaims.filter((claim) =>
```

jrhender (commented):
Could you also add a test that covers this scenario please? In addition to covering the case for future refactors, we can document there why this case needs to be covered.

JGiter (approved):

nichonien:
> Could you also add a test that covers this scenario please? In addition to covering the case for future refactors, we can document there why this case needs to be covered.

@jrhender This is just to cut down the claims to have only acceptedRoles before verification. Can you suggest what this test could do / have?

jrhender:
> > Could you also add a test that covers this scenario please? In addition to covering the case for future refactors, we can document there why this case needs to be covered.
> 
> @jrhender This is just to cut down the claims to have only acceptedRoles before verification. Can you suggest what this test could do / have?

I am thinking that the test would return a `this.credentialResolver.eip191JwtsOf` and confirm that the behaviour is still correct (admittedly, I'm not sure how to asset "correct" behaviour here). The test case could be called something along the lines of "LoginService should function even if user has unverified roles".

jrhender:
@nichonien When you have a moment, could you explain the connection between role acceptance and the role having `claimData.claimType`? I would think that even unaccepted roles would have `claimType`

nichonien:
> @nichonien When you have a moment, could you explain the connection between role acceptance and the role having `claimData.claimType`? I would think that even unaccepted roles would have `claimType`

@jrhender While authorising, a system expects some role credential from a user ( these are accepted roles, defined by system ). So with this PR we are filtering user's role credential (claimData.claimType ) matches accepted roles and if true then only those role credential should be verified.

jrhender (changes requested):
Ok, I think I understand the PR better now. I think we should not merge this. I think the intention with `acceptedRoles` is a user can only get the authentication token if they have at least one of the accepted roles. It is not intended to filter out the roles that are to be verified or included in the returned token.
For instance, I think for the did-auth-proxy deployment for Project EDGE, we use an empty `acceptedRoles` array. -> Is this right @artursudnik ?

artursudnik:
No, `acceptedRoles` in `did-auth-proxy` needs to be populated with at least one role. If any DID could get an access token, putting this proxy in front of a REST API would not add any protection without modifying logic within this REST API backend.

jrhender:
> However, I see a problem in the fact that after DID is authenticated by PDA, the backend gets a list of all roles. I would filter out all roles except listed in acceptedRoles

I think that's in essence what this PR is doing, right? @nichonien 

I'm open to this change, but I think this is significant breaking change because any existing app using passport-did-auth would have a different set of roles returned in the token if they don't update their `acceptedRoles` config. Am I understanding this correctly @artursudnik @nichonien ?

nichonien:
> However, I see a problem in the fact that after DID is authenticated by PDA, the backend gets a list of all roles. I would filter out all roles except listed in `acceptedRoles`

@artursudnik This PR addresses this problem only.

artursudnik:
Yes, this would be a breaking change. I can imagine an implementation where the `acceptedRoles` is empty, any role is accepted but roles are verified somewhere else after the token is presented to another system for example.

jrhender:
I think another arguable downside to this change is that verifiers (users of passport-did-auth) would now need to reconfigure passport-did-auth every time they want to verify a new role. Though maybe this is "standard" in SSI: i.e. if a verifier wants a new credential to be presented from a holder, they need to add it to their presentation definition.

What are your thoughts on the above @nichonien @artursudnik @mhrsntrk ? ☝️

artursudnik:
But, actually, any system having passport-did-auth integrated into it needs to also have roles to be verified somewhere. Otherwise, anyone could access it having any random role.

nichonien:
>I think another arguable downside to this change is that verifiers (users of passport-did-auth) would now need to reconfigure passport-did-auth every time they want to verify a new role.

This was the case always AFAIK, and this to me looks the authentic way to update accepted roles. 

@jrhender Can we confirm with projects that are using `passport-dud-auth`, if they are again verifying the roles (other than the accepted roles) somewhere in there system for any use case. If not, then this really isn't a breaking change.

lib/LoginStrategy.ts
```diff
@@ -137,6 +140,9 @@ export class LoginStrategy extends BaseStrategy {
     this.numberOfBlocksBack = numberOfBlocksBack;
     this.jwtSecret = jwtSecret;
     this.acceptedRoles = new Set(acceptedRoles);
+    if (includeAllRoles) {
```
artursudnik, line 143: is this "if" necessary here?

nichonien: Yeah, this is an optional parameter.

lib/LoginStrategy.ts
```diff
@@ -266,8 +272,13 @@ export class LoginStrategy extends BaseStrategy {
         this.cacheServerClient?.address === userAddress
           ? []
           : await this.credentialResolver.eip191JwtsOf(userDid);
+      const roleClaims = this.includeAllRoles
+        ? userClaims
+        : userClaims.filter((claim) =>
+            this.acceptedRoles.has(claim.payload.claimData.claimType)
```
jrhender, line 278: I think we should check that `claimType` is available in the payload to available null-reference error. Even if we're sure of the behaviour of `eip191JwtsOf` currently (i.e. we know that it always return JWTs with `claimType`) I think it's good to have this check in case the behaviour changes.
```suggestion
        : userClaims.filter((claim) => {
            const claimType = claim?.payload?.claimData?.claimType;
            return claimType && this.acceptedRoles.has(claimType);
          }
```

jrhender (commented):
@nichonien I think it's looking good. Could you add a few test cases for the scenarios that being added here please? I'm thinking it would be good to, at a minimum, have tests for:
1. verified roles only include roles in `acceptedRoles`
2. `includeAllRoles` overrides `acceptedRoles` and includes all roles
I think you might be able to use this test as inspiration? https://github.com/energywebfoundation/passport-did-auth/blob/6caa1aa0c77b84ec0225ff3155dd0e081cb3a65d/test/LoginStrategy.test.ts#L353

Also, one of these commits should be marked as `feat` with "BREAKING CHANGE" to indicate that the behaviour of passport-did-auth will be changing in a breaking manner

JGiter (approved):

test/LoginStrategy.test.ts
```diff
@@ -350,6 +350,186 @@ it('Should support old format did for off chain claims', async () => {
...
+    claimData: {
+      blockNumber: 4242,
+    },
+    sub: '',
+  };
+
+  if (!token) {
+    expect(false).toBeTruthy();
+    return;
+  }
```
jrhender, line 518: Similar to previous comment. Should just assert against the token, IMO

test/LoginStrategy.test.ts
```diff
@@ -350,6 +350,186 @@ it('Should support old format did for off chain claims', async () => {
...
+  jest
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    .spyOn((loginStrategy as any).credentialResolver, 'eip191JwtsOf')
+    .mockResolvedValueOnce([claim1, claim2]);
+
+  const consoleListener = jest.spyOn(console, 'log');
+  await loginStrategy?.validate(token, payload, () => {
+    expect(consoleListener).toBeCalledWith(
+      'includeAllRoles: false, verifying only accepted roles'
+    );
```
jrhender, line 467: I think using `consoleListener` is an okay solution, but not great because the log could lie (e.g. it could be in the wrong place). I noticed that the `done` callback param in the `validate` function has a `user` param. https://github.com/energywebfoundation/passport-did-auth/blob/fe0f171aa4b1fcef4cd151819bb3d3d2dd368047/lib/LoginStrategy.ts#L211
Could we assert against that `user` param perhaps? It seems to contain the verified roles result? https://github.com/energywebfoundation/passport-did-auth/blob/fe0f171aa4b1fcef4cd151819bb3d3d2dd368047/lib/LoginStrategy.ts#L295

jrhender (commented):
Great start to the tests but I added some suggestions as well

test/LoginStrategy.test.ts
jrhender: @nichonien what do you mean by "nested scenarios"?

nichonien: I mean scenarios where we need to mock calls that cannot be directly accessed (for ex. If we want to mock getRoleDeifition in IssuerVerification) or are purely primitive types.

jrhender: Yes, I thought that this would be possible:
> So for includeAllRoles to be asserted against user param it should be some actual role claims to be issued and get verified.

But you'd saying that the above is quite challenging, because we can't mock or spy on primitive types. Is this right?

From what I can see, if `ClaimVerifier` was mocked, then the role claim wouldn't have actually be issued & verified as `verifier.getVerifiedRoles()` could be mocked. Does this sound right or not really?

nichonien: > But you'd saying that the above is quite challenging, because we can't mock or spy on primitive types. Is this right?
Yes, mocking is hard. With actual claims, test might take time since it would be validating them through `Volta` test network or may get  in pending state.

`ClaimVerifier` can be mocked if it is instantiated independently (not for LoginStrategy instance as verifier is primitive type) and I was not able to mock `verifier.getVerifiedRoles()`. Also if you mock `verifier.getVerifiedRoles()` and get some expected output aligning to `includeAllRoles` it doesn't seem to be a fit test as it would never allow the validation ever to happen (filtering based on `includeAllRoles` param). Does this answer you concern, or I didn't get it actually?

jrhender: Ah I see, the `ClaimVerifier` instantiation is internal to `LoginStrategy`. Yes, that does make mocking it challenging/unfeasible. Ok, I propose that we leave it as is for now and come back to improve the tests in the future if necessary.

jrhender (approved):
Thanks for all of your work on this @nichonien

PR #332, opened 2022-09-22
fix: fixing a bug when DIDs with no roles are always accepted

artursudnik:


nichonien (approved):

nichonien:
@artursudnik Looks like the test are not finishing (hangs up), I tried running the workflow again and it's the same case.

artursudnik:
I see tests are failing in a very strange way so that they hang and then time out.

PR #334, opened 2022-10-03
feat: add cache to resolvers

nichonien:


jrhender:
@nichonien as discussed previously, I think we need some kind of cache refresh/TTL strategy, otherwise passport-did-auth could return inaccurate results. Previously, we discussed potentially have the cache be "per-authentication" request. What do you think of this approach?

nichonien:
> @nichonien as discussed previously, I think we need some kind of cache refresh/TTL strategy, otherwise passport-did-auth could return inaccurate results. Previously, we discussed potentially have the cache be "per-authentication" request. What do you think of this approach?

I think the current approach might not have this issue, as for every issuer verification call it sets a new cache and this might provide the latest claim status for authentication. https://github.com/energywebfoundation/ew-credentials/blob/298ebc7d46b92e55935df8817353ac4e79f59d15/packages/vc-verification/src/verifier/issuer-verification.ts#L83. Do you think there could be a different case that might not be covered in this approach?

JGiter (approved):

JGiter (approved):

jrhender (approved):
Looks good, but don't we need to also create the caches and pass them in where appropriate?

nichonien:
> Looks good, but don't we need to also create the caches and pass them in where appropriate?

The issuer-verification classes has all the code to create new caches and pass them internally (already done in the ew-credentials PR). In PDA only different implementation of resolvers are passed (resolution from SSI-Hub) and this had to be updated to accommodate caching.

PR #339, opened 2022-10-13
feat: return reason for failed authorisation

nichonien:


JGiter (changes requested):
If possible please replace error strings with error classes from vc-verification

lib/LoginStrategy.ts
```diff
@@ -287,19 +278,44 @@ export class LoginStrategy extends BaseStrategy {
...
+            (verifiedRole) => verifiedRole.namespace === role
+          );
+          if (!acceptedRole) {
+            invalidRoles.push({
+              name: role,
+              namespace: role,
+              error: 'Role credential not found for the user.',
+            });
+            userValidationStatus = false;
+          } else if (acceptedRole.error) {
```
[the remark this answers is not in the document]
nichonien: I think the naming of `acceptedRole` is wrong, it is not related to `acceptedRoles` but it is a declaration to process intermediate values. I will change it.

lib/ClaimVerifier.ts
```diff
@@ -48,6 +49,7 @@ export class ClaimVerifier {
   private async verifyRole(credential: Required<RoleEIP191JWT>): Promise<{
     name: IRoleDefinitionV2['roleName'];
     namespace: RoleEIP191JWT['payload']['claimData']['claimType'];
+    error: string;
```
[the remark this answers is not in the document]
nichonien: Yeah, it eases the validation based on null. Like this https://github.com/energywebfoundation/passport-did-auth/blob/2f0d23d8d7bb3754cdad6827cd2d8c96d6a8a2d6/lib/LoginStrategy.ts#L298

lib/LoginStrategy.ts
```diff
@@ -422,4 +401,41 @@ export class LoginStrategy extends BaseStrategy {
 
     return true;
   }
+
+  /**
+   * Filters verified roles into valid and invalid roles and sets validation status
+   * @param uniqueRoles verified roles
+   * @returns
+   */
+  filterUserRoleStatus(uniqueRoles: RoleStatus[]): {
```
JGiter, line 410: Is this method needs to be exposed?

lib/LoginStrategy.ts
```diff
@@ -422,4 +401,41 @@ export class LoginStrategy extends BaseStrategy {
 
     return true;
   }
+
+  /**
+   * Filters verified roles into valid and invalid roles and sets validation status
```
JGiter, line 406: I beleive that `invalid` isn't appropriate in this case. May be `notSaticfiedRoles`. Also `valid` isn't needed, because user most often need to know which requirements he yet need to satisfy

lib/ClaimVerifier.ts
```diff
@@ -32,6 +33,7 @@ export class ClaimVerifier {
     const filteredRoles = roles.filter(Boolean) as {
       name: IRoleDefinitionV2['roleName'];
       namespace: RoleEIP191JWT['payload']['claimData']['claimType'];
+      error: string;
```
JGiter, line 36: Its still looks odd to me that `getVerifiedRoles` returns not verified role

JGiter (changes requested):

lib/LoginStrategy.ts
nichonien: No, will make it private .

lib/LoginStrategy.ts
nichonien: Valid credentials are needed, since we display the verified roles once authorised in `ram-client-examples`. PDA has always returned verified roles.

lib/ClaimVerifier.ts
nichonien: 1. To me verification is about ensuring the data came from the right channel and its integrity is still persisted. Validity is solely based on business use case and it is next step after verification. For example - I was issued driving license which was valid for 10 years. Even after expiration I can use it to prove my age at a Bar but I cannot drive without renewal.
2. `getVerifiedRoles` is solely responsible for verification and if one wants to also know the invalid roles (expired or revoked) then it will come from this call itself. 

Maybe we can segregate the valid and revoked / expired role here itself so that there is no confusion.

lib/LoginStrategy.ts
JGiter: Ok, thanks. Just to clarify, is it assumed that `validate` returns only those user roles which present in accepted?

nichonien: It depends on another parameter `includeAllRoles` - https://github.com/energywebfoundation/passport-did-auth/blob/6028f44f026b615e93db5095b4c530c7d1d5c1b4/lib/LoginStrategy.ts#L66.

So, if this is true it will return all verified roles and if not then only the accepted roles.

JGiter: But isn't `filterUserRoleStatus` rerurns only accepted roles regardless of `includeAllRoles`?

JGiter: I think we just shouldn't call `filterUsrRoleStatus`  when `includeAllRoles`

nichonien: @JGiter Thinking more on the above, what should we do of revoked or expired roles if the `includeAllRoles` is true? Should it be separated in a different list?

jrhender: @JGiter @nichonien where is `filterUserRoleStatus`? I don't see it in this code

> Thinking more on the above, what should we do of revoked or expired roles if the includeAllRoles is true? Should it be separated in a different list?

@nichonien does your question about still stand? To me, they can just be roles in the `userRoles` array that are marked as `revoked` or `expired`

lib/LoginStrategy.ts
```diff
@@ -406,4 +413,38 @@ export class LoginStrategy extends BaseStrategy {
...
+            name: role,
+            namespace: role,
+            error: 'Role credential not found for the user.',
+          });
+          authorisationStatus = false;
+        } else if (userRole.error) {
+          authorisationStatus = false;
+        }
+      });
+    }
```
jrhender, line 444: @nichonien Is this is a change in the behaviour from before?
I think the previous code (below) passed if **any** of the roles from `acceptedRoles` where verified and present. I think the new code requires that  **all** accepted roles are present?
If this is indeed a change, was this change intended?
```
this.acceptedRoles.size > 0 &&
!uniqueRoles.some(({ namespace }) => {
          return this.acceptedRoles.has(namespace);
})
```

jrhender (changes requested):
The change in behaviour in `validateAcceptedRoles` from any to all is a potential issue

lib/LoginStrategy.ts
nichonien: > @JGiter @nichonien where is filterUserRoleStatus? I don't see it in this code

I renamed it since we decided to return just one array of roles.

lib/LoginStrategy.ts
nichonien: @jrhender No, this is a mistake. I will correct it.

jrhender: Ok, great. Maybe we can add a test for this case (should be authorized when only one of the acceptedRoles is verified and other is not verified) either in this PR or a separate one?

nichonien: Adding a test is not possible, since mocking `ClaimVerifier` is not feasible. Our earlier discussion - https://github.com/energywebfoundation/passport-did-auth/pull/331#discussion_r985457467

lib/ClaimVerifier.ts
```diff
@@ -32,6 +34,7 @@ export class ClaimVerifier {
     const filteredRoles = roles.filter(Boolean) as {
```
JGiter, line 34: Why to cast to `{name; namespace; status}`?

JGiter (approved):

lib/ClaimVerifier.ts
nichonien: @JGiter I tried removing this code and it results in some abnormalities in returned `verifiedRoles`. Maybe I give it some more time and remove this in a separate PR? So that the UI in `iam-client-examples` could pick this informative rejection changes ASAP. @jrhender what are your thoughts?

lib/ClaimVerifier.ts
```diff
@@ -63,8 +67,16 @@ export class ClaimVerifier {
...
-      return null;
+    const roleStatus = await this.checkRoleStatus(credential.payload);
+    if (!roleStatus.verified) {
+      return {
+        name: role.roleName,
+        namespace: credentialClaimData.claimType,
+        status:
+          roleStatus.error === CredentialRevoked
+            ? RoleCredentialStatus.REVOKED
+            : RoleCredentialStatus.EXPIRED,
```
jrhender, line 78: The code as-is is a bit brittle because if a new error is added but this code isn't changed, then `RoleCredentialStatus.EXPIRED` would be returned incorrectly. What about this change below?
```suggestion
      const errorMessageMap = {
        CredentialExpired: RoleCredentialStatus.EXPIRED,
        CredentialRevoked: RoleCredentialStatus.REVOKED
      }
      return {
        name: role.roleName,
        namespace: credentialClaimData.claimType,
        status: errorMessageMap[roleStatus.error] ?? RoleCredentialStatus.UNKNOWN
```

lib/ClaimVerifier.ts
jrhender: > Maybe I give it some more time and remove this in a separate PR

I agree, better to do it in a separate PR

lib/ClaimVerifier.ts
nichonien: @jrhender I agree with you, but then if a new `error` is added then we would need to update error mapping and add error to `RoleCredentialStatus` as well to make it meaningful. So the code would be needed to update anyhow.

jrhender (approved):
Looks good, thanks!

PR #342, opened 2022-11-04
build: update dependencies

nichonien:


jrhender (approved):

jrhender (changes requested):
I think `chore` is the better commit message prefix here: https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/

> chore – changes that do not relate to a fix or feature and don't modify src or test files (for example updating dependencies)

jrhender (approved):
Thanks, looks good!

PR #351, opened 2022-12-06
chore: revert typescript to v4.7.4

nichonien:


jrhender:
@nichonien why the reversion? Seems to me that it's generally better to have newer version

nichonien:
> @nichonien why the reversion? Seems to me that it's generally better to have newer version

The upgrade was made as part of `depandabot` PR, it runs into version compatibility errors with `typedoc` and `ethers`. Considering previous experiences, it is better to upgrade all these together and should be done in `ew-did-reg` first.

jrhender:
Ok, so this revision is a preemptive change to reduce the likelihood of future errors? Or are there errors that you are currently experiencing?

jrhender (approved):
Seems okay if required.

nichonien:
> Ok, so this revision is a preemptive change to reduce the likelihood of future errors? Or are there errors that you are currently experiencing?

Both, but the urgency is for the current issues. I will be upgrading these dependency in `ew-did-reg` and then in PDA, so that import doesn't break anything (mostly happens with `ethers`).

PR #360, opened 2023-01-17
fix: downgrade didstore to v0.7.1-alpha.816.0

nichonien:
This PR downgrades `DidStore` to `v0.7.1-alpha.816.0`. This makes `passport-did-auth` compatible to the current version of `ew-credentials` which uses the IPFS Gateway instead of Cluster.


lib/LoginStrategy.ts
```diff
@@ -48,6 +49,7 @@ export interface LoginStrategyOptions extends StrategyOptions {
   }[];
   didContractAddress: string;
   ensRegistryAddress: string;
+  ipfsUrl?: string;
```
jrhender, line 52: Rather than a Url, would be better to accept a config object that takes url and headers. This is what did-auth-proxy is providing I believe: https://github.com/energywebfoundation/did-auth-proxy/blob/b20ddef5f37a3b5cf2a9fb597b1cdcef4d089020/authorization-server/src/modules/auth/strategies/auth.strategy.ts#L84

nichonien: @jrhender `did-auth-proxy` is an app and it is efficient and best to use `ConfigService` from `nestjs` for all the configuration parameters being consumed. Using `ipfsUrl` here keeps it simple and avoids unnecessary processing, also very less work to revert back to `Cluster` version implementation. What are your thoughts?

jrhender: `ipfsUrl` as string does keep it simple but it doesn't reflect the reality that an object is needed to use Infura API as the API requires authentication. Was `ipfsUrl` a string in previous versions of PDA? If so we can keep it as a string and improve in a separate PR

nichonien: Yes, it was string in previous versions - https://github.com/energywebfoundation/passport-did-auth/blob/1a291c2db1c197d2230a4f1a662623a34fdac067/lib/LoginStrategy.ts#L52.

jrhender: Ok, sounds good, let's keep as is then

jrhender (approved):

PR #365, opened 2023-02-17
feat: add support for siwe

nichonien:


lib/LoginStrategy.types.ts
```diff
@@ -26,6 +26,20 @@ export interface ITokenPayload {
   iss: string;
 }
 
+export interface ISiweMessagePayload {
```
jrhender, line 29: I think this can be replaced by the type from the `siwe` package: https://github.com/spruceid/siwe/blob/053f16bf7dce3ff587e00bc4d3754a29ee9c26e6/packages/siwe/lib/client.ts#L20

jrhender (commented):
Looks like a good start, some initial comments

lib/LoginStrategy.types.ts
nichonien: @jrhender That's not a `type`, it is the main `Class`. I looked at their code to see if they have any interface for these attributes.

lib/LoginStrategy.ts
```diff
@@ -424,6 +486,28 @@ export class LoginStrategy extends BaseStrategy {
     return true;
   }
 
+  isSiweMessagePayload(payload: unknown): payload is ISiweMessagePayload {
+    if (!payload) return false;
+    if (typeof payload !== 'object') return false;
+    const payloadKeys = Object.keys(payload);
+    if (
+      !hasIn(payload, 'domain') &&
```
JGiter, line 494: All these fields are required
```suggestion
      !hasIn(payload, 'domain') ||
      !hasIn(payload, 'nonce') ||
      !hasIn(payload, 'uri') ||
      !hasIn(payload, 'address') ||
      !hasIn(payload, 'chainId') ||
      !hasIn(payload, 'version')
```

lib/LoginStrategy.types.ts
JGiter: Why you can not use class?
If class is indeed inapropriate you can try siwe.SiweMessage as type or
```
import type {SiweMessage} from 'siwe`;
```

lib/LoginStrategy.ts
```diff
@@ -424,6 +486,29 @@ export class LoginStrategy extends BaseStrategy {
...
+    ) {
+      return false;
+    }
+    if (
+      typeof payload['domain'] !== 'string' ||
+      typeof payload['nonce'] !== 'string' ||
+      typeof payload['uri'] !== 'string'
+    ) {
+      return false;
+    }
```
jrhender, line 508: I think the `hasIn` checks are redundant here because if the key isn't on the object then the type will be `undefined` which will be caught by the type checks.
I also think that by casting the property to the type we want, we can have compilation errors in the future if we upgrade `siwe` package and the `SiweMessage` class changes.
```suggestion
    const castedPayload = payload as SiweMessage;
    if (
      typeof castedPayload.domain !== 'string' ||
      typeof castedPayload.nonce !== 'string' ||
      typeof castedPayload.uri !== 'string' ||
      typeof castedPayload.address !== 'string'
    ) {
      return false;
    }
```

lib/LoginStrategy.types.ts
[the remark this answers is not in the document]
nichonien: Missed to add the file to commit.

test/LoginStrategy.siwetoken.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    ensRegistry.address
+  );
+  const wrongPayload = { ...sampleSiwePayload };
+  delete wrongPayload.nonce;
+  const consoleListener = jest.spyOn(console, 'log');
+  await loginStrategy?.validate(token, wrongPayload, () => {
+    expect(consoleListener).toBeCalledWith('Token payload is not valid');
+  });
+  const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+  await delay(10000);
```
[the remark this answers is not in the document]
nichonien: Forgot to remove after test were successful.

lib/LoginStrategy.ts
```diff
@@ -32,6 +32,9 @@ import {
   VerificationResult,
 } from '@energyweb/vc-verification';
 import { StatusListEntryVerification } from '@ew-did-registry/revocation';
+import { SiweMessage, SiweResponse } from 'siwe';
+import type { SiweMessage as SiweMessagePayload } from 'siwe';
+import { hasIn } from 'lodash';
```
jrhender, line 37: I think `lodash` isn't in the dependencies listed in `package.json`, so this is being accessed without being listed, which should be avoided.
From what I can tell `hasIn`, is most useful if there is a path to traverse which isn't the case at the moment, so I think we can just just straight object indexing

jrhender (commented):
A couple more comments

lib/LoginStrategy.ts
```diff
@@ -256,6 +275,49 @@ export class LoginStrategy extends BaseStrategy {
...
+        nonce: payload.nonce,
+        time: payload.issuedAt,
+      });
+    } catch (error) {
+      const err = error as SiweResponse;
+      const errorMessage = err.error?.type;
+      return done(new Error(errorMessage));
+    }
+    return this.getVerifiedAcceptedRoles(userDid, done);
+  }
```
jrhender, line 308: This verification should also check the `uri` value of the payload as I think this check is not being performed by the `siwe` library, I believe.
I think the expected `uri` should be set in the `LoginStrategy` constructor.
This check is important in order to help prevent replay attacks.

jrhender (changes requested):
`uri` check is missing

JGiter (commented):
@nichonien `BaseStrategy.extractToken` also need to be changed. It need to get `message` and `signature` from request and pass them to `LoginStrategy.validate`. This change is another reason to have two passport strategies - one for each authentication method

jrhender:
@JGiter 
> This change is another reason to have two passport strategies - one for each authentication method

If there were two passport strategies, would that complicate the integration in ssi-hub and did-auth-proxy? I guess it wouldn't, assuming that there is a different endpoint path for each strategy?

JGiter:
@jrhender I don't expect complications caused by having strategies per authentication method. From my point of view we need to apply AuthGuard('identityToken') to `/login` route and AuthGuard('siwe') to `/login/siwe/verify`

nichonien:
> @nichonien `BaseStrategy.extractToken` also need to be changed. It need to get `message` and `signature` from request and pass them to `LoginStrategy.validate`. This change is another reason to have two passport strategies - one for each authentication method

@JGiter I'm not really following, do you think the current `extractToken` would be unable to extract `signature`? It is because of the naming convention? I think did-auth-proxy or any other module would generally send signatures in form of `token` and the current authenticate method would be able to pick it up. @artursudnik any thoughts on this?

jrhender:
Hmm, I think this is like a tricky situation, unless I'm missing something.
I realize now that top-level logic for the strategy is in the `authenticate` method of the `BaseStrategy`, which is essentially:

```javascript
authenticate(req: Request): void {
    const token = this.extractToken(req);
    if (!token) {
      return this.fail('Missing credentials', 400);
    }
    const tokenPayload = this.decodeToken(token);
    this.validate(token, tokenPayload, verified);
}
```

So for the above to work for siwe, `token` needs to be the `signature` and `tokenPayload` needs to be `message`, taken from the body of the `req` (looking at the current PR in ssi-hub): https://github.com/energywebfoundation/ssi-hub/pull/582
This seems really challenging to do, because it would mean that `decodeToken` would be a mapping from `signature` to `message`, which doesn't seem possible. I think the core problem is that `BaseStrategy` is currently designed with JWTs in mind, not a SIWE request which has signature and message in two separate props.

Probably the best way would be to redesign the strategy hierarchy/composition to be more flexible and single-responsible. But I don't think we have time for that, unless you guys see clearly what the better design is.
I think a faster way could be to modify the `authenticate` logic. Maybe something like:

```javascript
authenticate(req: Request): void {
    const token = this.extractToken(req); // expect that it is null if not a JWT
    const siweObject = this.extractSiwe(req); // expect that it is null if message and signature can't be found
    if (token) {
      const tokenPayload = this.decodeToken(token);
      this.validate(token, tokenPayload, verified);
    }
    else if (siweObject) {
      this.validate(siweObject.message, siweObject.tokenPayload, verified);
    }
    else {
      return this.fail('Missing credentials', 400);
    }
}
```
What do you think of the above @JGiter @nichonien ?

jrhender:
Even if we don't have a perfect implementation to start, getting a working implementation out would be good so that we can start integrating all the way up to the frontends ASAP.

lib/LoginStrategy.ts
```diff
@@ -413,6 +425,27 @@ export class LoginStrategy extends BaseStrategy {
     );
   }
 
+  /**
+   * extracts siwe message and signature either from request body or query
+   * @param req
+   * @returns {string} siwe object - message and signature
+   */
+  extractSiwe(req: Request): SiweMessageReqPayload {
+    if (req.body.identity)
```
JGiter, line 434: Do you recall why message could be in `indentity` rather then in request body? Is it for legacy reason?

lib/LoginStrategy.types.ts
```diff
@@ -26,18 +27,9 @@ export interface ITokenPayload {
...
-  uri: string;
-  version: string;
-  chainId: number;
-  nonce: string;
-  issuedAt?: string;
-  expirationTime?: string;
-  notBefore?: string;
-  requestId?: string;
+export interface SiweMessageReqPayload {
+  signature: string | null;
```
JGiter, line 31: Both these fields should be required

lib/LoginStrategy.types.ts
```diff
@@ -1,5 +1,6 @@
 import { IRoleDefinitionV2 } from '@energyweb/credential-governance';
 import { VerifiablePresentation } from '@ew-did-registry/credentials-interface';
+import type { SiweMessage as SiweMessagePayload } from 'siwe';
```
JGiter, line 3: what is the reason of creating alias?

lib/LoginStrategy.ts
jrhender: I think for the old authentication protocol, we accept an object and expect the JWT to be on the `identity` prop or the prop that is specified during configuration. However, I think for siwe, we don't need to look for the `identity` prop and can directly look for `message` and `signature` in the request body.

lib/LoginStrategy.ts
```diff
@@ -351,6 +425,27 @@ export class LoginStrategy extends BaseStrategy {
     );
   }
 
+  /**
+   * extracts siwe message and signature either from request body or query
+   * @param req
+   * @returns {string} siwe object - message and signature
+   */
+  extractSiwe(req: Request): SiweMessageReqPayload {
```
JGiter, line 433: To be consistent with `extractToken` this method should return `SiweMessageReqPayload | null`

lib/LoginStrategy.types.ts
nichonien: No specific reason, just to avoid confusion for the use case.

lib/LoginStrategy.ts
```diff
@@ -351,6 +425,27 @@ export class LoginStrategy extends BaseStrategy {
...
+    if (req.body.identity)
+      return {
+        signature:
+          lookup(req.body.identity, 'signature') ||
+          lookup(req.query, 'signature'),
+        message: req.body.message,
```
JGiter, line 439: The `req.body.message` is string. It should be JSON parsed before return

JGiter (changes requested):

lib/LoginStrategy.ts
```diff
@@ -351,6 +425,27 @@ export class LoginStrategy extends BaseStrategy {
...
+      };
+
+    return {
+      signature:
+        lookup(req.body, 'signature') || lookup(req.query, 'signature'),
+      message: req.body.message,
+    };
+  }
```
jrhender, line 447: It's related to @JGiter 's comment about `identity`, but I think this implementation is a bit overcomplicated.
To me, it makes sense to have the properties on `SiweMessageReqPayload` be mandatory but return `null` if it can't be extracted (this is inline with how `extractToken` is implemented currently).

I also don't really see the value of `lookup`, but maybe I'm missing something there.

```suggestion
  extractSiwe(req: Request): SiweMessageReqPayload || null {
    if (req.body.message && req.body.signature) {
      return {
        signature: req.body.signature,
        message: req.body.message
      };
    }
    else {
     return null;
    }
  }
```

lib/LoginStrategy.ts
jrhender, line 433: Can be shortened by removing `Message`, IMO
```suggestion
  extractSiwe(req: Request): SiweReqPayload {
```

lib/LoginStrategy.ts
JGiter: Actually it is not in JSON, but plain text. We need to parse it line by line https://docs.login.xyz/sign-in-with-ethereum/quickstart-guide/creating-siwe-messages

JGiter: perhaps this can help https://www.npmjs.com/package/properties-file

artursudnik:
>  I think did-auth-proxy or any other module would generally send signatures in form of token and the current authenticate method would be able to pick it up. @artursudnik any thoughts on this?

I think that clients will need to adjust to strategies and strategies should comply with the standards/conventions as much as possible.

lib/LoginStrategy.ts
jrhender: Maybe the best way to do this is to reuse the `SiweMessage` constructor: https://github.com/spruceid/siwe/blob/053f16bf7dce3ff587e00bc4d3754a29ee9c26e6/packages/siwe/lib/client.ts#L62
```suggestion
        message: new SiweMessage(req.body.message),
```

lib/BaseStrategy.ts
```diff
@@ -80,7 +80,7 @@ export abstract class BaseStrategy extends Strategy {
     if (token) {
       const tokenPayload = this.decodeToken(token);
       this.validate(token, tokenPayload, verified);
-    } else if (siweObject.signature && siweObject.message) {
+    } else if (siweObject && siweObject.signature && siweObject.message) {
```
JGiter, line 83: If `siweObject` is not null, then it has `signature` and `message`
```suggestion
    } else if (siweObject) {
```

JGiter (changes requested):

lib/LoginStrategy.types.ts
```diff
@@ -26,6 +27,11 @@ export interface ITokenPayload {
   iss: string;
 }
 
+export interface SiweReqPayload {
```
JGiter, line 30: Please export it

lib/LoginStrategy.ts
```diff
@@ -351,6 +425,22 @@ export class LoginStrategy extends BaseStrategy {
     );
   }
 
+  /**
+   * extracts siwe message and signature either from request body or query
+   * @param req
+   * @returns {string} siwe object - message and signature
+   */
+  extractSiwe(req: Request): SiweReqPayload | null {
+    if (req.body.message && req.body.signature) {
```
jrhender, line 434: Looking at it again, I think having optional chaining would be useful here to avoid exceptions if the input is unexpected (e.g. if `req` doesn't have a `body` prop)
```suggestion
    if (req?.body?.message && req?.body?.signature) {
```

jrhender (approved):
Another small comment but I think it looks good to me if it works for @JGiter in ssi-hub https://github.com/energywebfoundation/ssi-hub/pull/582

JGiter:
Authentication works, but we also need to decide how to handle invalid Siwe message. Constructor of `SiweMessage` can fail. Perhaps we should wrap it in `try`. I would also like to provide meaninful error message, but `siwe` does not helps with this. Here is the example of error from `siwe`:
```
Invalid message: {"success":false,"state":103,"length":316,"matched":0,"maxMatched":280,"maxTreeDepth":19,"nodeHits":3523,"inputLength":316,"subBegin":0,"subEnd":316,"subLength":316}
```

JGiter:
@nichonien Could you also pass `InvalidSiweMessage` error in `done`, when message is invalid. Also since message is validated in SiweMessage constructor the `isSiweMessagePayload` is not needed

nichonien:
> @nichonien Could you also pass `InvalidSiweMessage` error in `done`, when message is invalid. Also since message is validated in SiweMessage constructor the `isSiweMessagePayload` is not needed

@JGiter The `isSiweMessagePayload` is needed to validate the payload so that we can redirect the request verification method i.e. `eip191` or `siwe`.

JGiter:
> @JGiter The `isSiweMessagePayload` is needed to validate the payload so that we can redirect the request verification method i.e. `eip191` or `siwe`.

You are right. A more precise check would be `message instanceof SiweMessage` though

lib/BaseStrategy.ts
```diff
@@ -58,10 +67,7 @@ export abstract class BaseStrategy extends Strategy {
    */
   authenticate(req: Request): void {
     const token = this.extractToken(req);
-    if (!token) {
-      return this.fail('Missing credentials', 400);
-    }
-    const tokenPayload = this.decodeToken(token);
+    const siweObject = this.extractSiwe(req);
```
jrhender, line 70: @JGiter regarding your comment about error handling, does it make sense to have a try/catch here? Is this what you have in mind?
```suggestion
    let siweObject: SiweReqPayload | null;
    try {
      siweObject = this.extractSiwe(req);
    } catch(err) {
      return this.fail(err as string)
    }
```

JGiter: Yes, but it would also be useful to pass error, which would allow to identify error reason

jrhender: Right, makes sense. How about this?
```suggestion
    let siweObject: SiweReqPayload | null;
    try {
      siweObject = this.extractSiwe(req);
    } catch(err) {
      return this.error(new Error('error while extracting SIWE object', { cause: err }));
    }
```
I'm trying to use the `cause` prop of Error:
- https://stackoverflow.com/a/69799384
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

JGiter: The message from `siwe` doesn't provide any useful details. The only thing we can do is to communicate that message is invalid in some way. To identify this case I propose to throw some easily recognizable errror type, say `InvalidSiweMessage`

jrhender:
@nichonien @JGiter what is remaining to do in this PR from your perspective?

README.md
```diff
@@ -68,18 +68,9 @@ end
 
 This class provides implementation for verification of issued roles credential. The verification ensures that the credentials were issued by the authorised issuers and are neither revoked nor expired. LoginStrategy can be configured to authenticate only EnergyWeb Roles.
 
-LoginStrategy relies on [`IssuerVerification`](https://github.com/energywebfoundation/ew-credentials/blob/develop/packages/vc-verification/src/verifier/issuer-verification.ts) internally for verification of the roles credential. 
+Too know more about `LoginStrategy` initialisation check [How to configure LoginStrategy](./configure-loginstrategy.md).
```
JGiter, line 71:
```suggestion
Too know more about `LoginStrategy` initialization check [How to configure LoginStrategy](./configure-loginstrategy.md).
```

JGiter (approved):
Thanks @nichonien. I think we can merge it now

PR #368, opened 2023-03-01
task/LoginStrategy login to ssi-hub with siwe

JGiter:


lib/cacheServerClient.ts
```diff
@@ -72,39 +72,68 @@ export class CacheServerClient {
         `DID Login Strategy is now logged into cache server after ${duration.toFixed()}ms`
       );
     });
-    const token = await retry.execute(async () => {
```
jrhender, line 75: We should keep this `retry`

lib/cacheServerClient.ts
```diff
@@ -73,38 +73,36 @@ export class CacheServerClient {
...
+
+      const signature = await this.signer.signMessage(message);
+
+      const { data } = await this.httpClient.post<{ token: string }>(
+        '/login/siwe/verify',
+        {
+          message,
+          signature,
+        }
+      );
```
artursudnik, line 98: I'd add a debug log line in case of error. It will reduce pain for someone who will troubleshoot :)

JGiter: I think the error should be logged already in retry https://github.com/energywebfoundation/passport-did-auth/blob/d17bd1fc3b6bc581b34acb775e34f76830d9f858/lib/cacheServerClient.ts#L60

jrhender (approved):
Looks good to me 👍

PR #370, opened 2023-03-13
fix: add chainName to LoginStrategyOptions and set on cache client co…

whitneypurdum:
- add chainName to LoginStrategyOptions interface
- use chainName in cacheServerClient to identify chain when creating did

### DIFF
--- a/lib/LoginStrategy.ts
+++ b/lib/LoginStrategy.ts
@@ -22,6 +22,7 @@ import {
   DomainReader,
   ResolverContractType,
   IRoleDefinitionV2,
+  EWC_CHAIN_ID,
 } from '@energyweb/credential-governance';
 import { knownResolvers } from './defaultConfig';
 import {
@@ -58,7 +59,6 @@ export interface LoginStrategyOptions extends StrategyOptions {
   jwtSecret?: string | Buffer;
   jwtSignOptions?: SignOptions;
   siweMessageUri?: string;
-  chainName?: string;
 }
 
 export class LoginStrategy extends BaseStrategy {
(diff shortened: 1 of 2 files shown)


lib/LoginStrategy.ts
```diff
@@ -302,7 +303,12 @@ export class LoginStrategy extends BaseStrategy {
         null
       );
     }
-    const userDid = this.didUnification(`did:ethr:${payload.address}`);
+    let userDid;
+    if ((await this.getChainName(this.provider)) === 'ewc') {
```
jrhender, line 307: Better to rely on the chainId of the SIWE message I think. It is more intuitive as it allows the DID to be generated solely from the message.
```suggestion
    if (payload.chainId === EWC_CHAIN_ID) {
```
If above is used, `getChainName` can also be removed

jrhender:
@nichonien how hard would it be to add a test for this case?

jrhender:
@nichonien I think a test would involve adjusting the SIWE payload to have the EWC chain ID https://github.com/energywebfoundation/passport-did-auth/blob/develop/test/LoginStrategy.siwetoken.test.ts#LL40C14-L40

jrhender (approved):
@nichonien perhaps we can merge this as is and then add a test in a separate PR? How does that sound?

nichonien:
> @nichonien I think a test would involve adjusting the SIWE payload to have the EWC chain ID https://github.com/energywebfoundation/passport-did-auth/blob/develop/test/LoginStrategy.siwetoken.test.ts#LL40C14-L40

Yes, I am trying to obtain a `siwe` signature with the `message` in order to add it to tests.

nichonien:
> @nichonien perhaps we can merge this as is and then add a test in a separate PR? How does that sound?

Yeah, sounds good. We can merge this PR now and I'll create another PR with tests once I get the required sample payload with `ewc` chainID

